PR #35, opened 2023-02-10
feat: Create Client with Client::default(); use default parameters

swnb:
Using `Client::default()` to create a Client does not set the default api_base, it will crash when sending request;
The purpose of this pr is to use `Client::default()` to create a Client and set the correct default parameters

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -14,7 +14,7 @@ use crate::{
     Completions, Embeddings, FineTunes, Models,
 };
 
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 /// Client is a container for api key, base url, organization id, and backoff
 /// configuration used to make API calls.
 pub struct Client {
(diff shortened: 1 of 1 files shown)


async-openai/src/client.rs
```diff
@@ -29,15 +29,23 @@ pub const API_BASE: &str = "https://api.openai.com/v1";
 /// Name for organization header
 pub const ORGANIZATION_HEADER: &str = "OpenAI-Organization";
 
-impl Client {
-    /// Create client with default [API_BASE] url and default API key from OPENAI_API_KEY env var
-    pub fn new() -> Self {
+impl Default for Client {
+    // Create client with default [API_BASE] url and default API key from OPENAI_API_KEY env var
```
64bit, line 33:
```suggestion
    /// Create client with default [API_BASE] url and default API key from OPENAI_API_KEY env var
```
Lets preserve the comment for [doc](https://doc.rust-lang.org/rust-by-example/meta/doc.html) as well

async-openai/src/client.rs
```diff
@@ -29,15 +29,23 @@ pub const API_BASE: &str = "https://api.openai.com/v1";
...
             api_key: std::env::var("OPENAI_API_KEY").unwrap_or_else(|_| "".to_string()),
-            ..Default::default()
+            org_id: Default::default(),
+            backoff: Default::default(),
         }
     }
+}
+
+impl Client {
+    /// Create client with default parameters
```
64bit, line 45:
```suggestion
    /// Create client with default [API_BASE] url and default API key from OPENAI_API_KEY env var
```
Preserve comment for doc.rs?

64bit:
Good catch @swnb thank you for the fix.  
I'll release this in `0.6.1`, do you mind taking the look at comments?

swnb:
thank you for your reply @64bit

64bit (approved):

64bit:
The release `0.6.1` with this change is published on crates.io 
Thank you

PR #46, opened 2023-03-03
ChatChoice finish_reason can be null

ming08108:
finish_reason can sometimes be null in CreateChatCompletionResponse.

### DIFF
--- a/async-openai/src/types/types.rs
+++ b/async-openai/src/types/types.rs
@@ -771,7 +771,7 @@ pub struct CreateChatCompletionRequest {
 pub struct ChatChoice {
     pub index: u32,
     pub message: ChatCompletionResponseMessage,
-    pub finish_reason: String,
+    pub finish_reason: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


64bit:
Hi @ming08108 , thank you for PR. 

Is this always the case or its intermittent? 

In my testing I did not observe that otherwise [examples/chat](https://github.com/64bit/async-openai/tree/main/examples/chat) would have crashed .

One exception is Chat API with SSE streaming, where I did notice it to be null, which is already an `Option<String>` here https://github.com/64bit/async-openai/blob/main/async-openai/src/types/types.rs#L803

jonaro00:
@64bit I am also getting null on this field. In my case, the chance seems to be around 1 in 15.

zurawiki:
+1 this is happening for me too

64bit:
aha, that's no good,  lets ship this

64bit (approved):

64bit:
Its released in `0.9.0`. Thank you everyone

Issue #47, opened 2023-03-03
Allow customization of internal HTTP client


jtaox:
Hello! I'm using your  library and I noticed that there's no option to customize the internal HTTP client used for API requests when calling Client::new(). This makes it impossible to implement a proxy for all HTTP requests.

I was wondering if it would be possible to add a customization option for the internal reqwest client, similar to what the [chatgpt-api](https://github.com/transitive-bullshit/chatgpt-api/blob/main/src/chatgpt-api.ts#LL61C26-L61C26) offer. This would allow for greater flexibility and more use cases for this library.

64bit:
Hi @jtaox ,  happy to know you found it useful. 

Can you elaborate more about the use case with custom reqwest/HTTP client? What customization would you like to do?

I havent had to use a proxy yet, so your use case can help me understand better.

TmLev:
@64bit sometimes you may have a misbehaving virtual machine that doesn't like certain IP addresses. In such case you can specify an HTTP proxy for `reqwest` to proxy all the requests.

jtaox:
@64bit  In China, some websites are banned by the GFW, making it impossible to access them directly. Therefore, it's necessary to use a foreign server as a proxy to access these websites

jtaox:
@64bit  @TmLev Thank you for your patient response. After carefully reviewing the request documentation at [proxies](https://docs.rs/reqwest/latest/reqwest/#proxies), I found that reqwest enables proxies by default, as long as the system environment variables are set correctly. Therefore, #48  does not need to be merged.

64bit:
Aha glad you found a solution :) 

Thank you for contributing and sharing about GFW.

I'll close this as it's resolved for you.

TmLev:
While proxies may be set up by the means of environment variables, supporting custom middleware is still a thing :)

For instance, there is a tracing middleware which enables the use of OpenTelemetry traces: https://crates.io/crates/reqwest-tracing

64bit:
Hi @TmLev , Your use case is different from the original need for it, but this issue is broad enough so I'll  re-open this.

For having middlewares on client side, I see more info on https://github.com/seanmonstar/reqwest/issues/155 and https://truelayer.com/blog/adding-middleware-support-to-rust-reqwest/

So until `reqwest` natively support it, PRs are welcome, if this changes the type signature then additional work required then #48?

TmLev:
[...]

PR #48, opened 2023-03-03
Allow customization of internal reqwest client

jtaox:
#47

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -2,6 +2,7 @@ use std::pin::Pin;
 
 use futures::{stream::StreamExt, Stream};
 use reqwest::header::HeaderMap;
+use reqwest::{Client as ReqwestClient};
 use reqwest_eventsource::{Event, EventSource, RequestBuilderExt};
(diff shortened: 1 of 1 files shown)


64bit:
Thank you for the PR. Happy to continue use case dicussion on #47 . 

For PR: What do you think about changing `self.reqwest_client` to  `self.reqwest_client.clone()` on every request? 
I'm not sure about how stateful the reqwest client is, and would past requests affect future requests on reusing client?

TmLev:
@jtaox while you're at it, use `reqwest_middleware::ClientWithMiddleware` https://docs.rs/reqwest-middleware/latest/reqwest_middleware/struct.ClientWithMiddleware.html. Will allow using custom middleware.

jtaox:
@64bit I think it is not necessary to clone the reqwest client again, because the client.rs file currently does not modify the state of the reqwest client. Moreover, in the request code [client.rs](https://github.com/seanmonstar/reqwest/blob/c2ac875c36b96adf92c4c2c58afa5b185f075e6c/src/async_impl/client.rs#L1485), there is already a clone of the client made internally on every request.

PR #50, opened 2023-03-04
feat: Add max_tokens for chat completion requests

### LINKED ISSUE #49: max_tokens and user fields for CreateChatCompletionRequestArgs
It's seem like `max_tokens` filed is missed in args builder, but it presented in documentation:
https://platform.openai.com/docs/api-reference/chat/create

zurawiki:
- Add the `max_tokens` field to the `CreateChatCompletionRequest` struct.
- Updated example to verify behavior

Test plan:
Run chat and chat-stream examples locally

Closes #49


64bit:
Thank you for PR @zurawiki , its seems like a day for multiple release. 
I'll add the `user` field missing as well before releasing new version

64bit (approved):

PR #52, opened 2023-03-05
Fix: invalid_request_error: None is not of type 'string' - 'user'

dmweis:
This Pr should fix `Error: invalid_request_error: None is not of type 'string' - 'user'` error from the API.

On latest version of the library my requests are failing with the mentioned error. I assume this is because serde json will be default populate the field with `"null"` and the API expects it to be missing if it's not set?

`skip_serializing_if ` is already implemented on a lot of other fields and I suspect that it might also need to be added to others


64bit:
Hi @dmweis , thank you for fixing the bug! 

I should have tested it before publishing... I'll get this shipped right away

64bit (approved):

64bit:
Its released in v0.9.2

dmweis:
> Hi @dmweis , thank you for fixing the bug!
> 
> I should have tested it before publishing... I'll get this shipped right away

Thanks!

PR #55, opened 2023-03-11
Make API types clonable

dmweis:
Hey!

I originally only wanted to make `Usage` clonable because I was trying to pass it around and realise you can't clone it.
But looking at it I noticed that a lot of the API types don't have clone so I added it because it makes working with them easier.
I think some of them could even do with copy since they can be bitwise copied but didn't want to go around randomly adding things that I didn't need yet.

Let me know what you think! Happy to remove a lot of these clones if if you think I put them on types that shouldn't have them!

### DIFF
--- a/async-openai/src/types/types.rs
+++ b/async-openai/src/types/types.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::error::OpenAIError;
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone)]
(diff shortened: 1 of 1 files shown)


64bit:
Hi @dmweis , I like it! Having deriving `Clone` is useful and moreover you implementing it for all of them makes library consistent.

64bit (approved):

dmweis:
Awesome! Thanks!

64bit:
Thank you for the PR, this is released in `v0.9.4`

PR #57, opened 2023-03-12
Added dotenv support for client creation

FloppyDisck:
Many projects prefer to use dotenv since it makes managing credentials much easier. So I added a simple function that handles this for you, all you need to do is run `Client::from_env()` after loading your `env` file.


64bit:
Hi @FloppyDisck,

Thank you for the PR and effort you put into it. 

I myself use dotenv, and I agree that managing credentials is much easier with it. 

These are few reasons I can think of why adding dependency and env vars is not desirable for maintainence of `async-openai` long term:

1. Current `async-openai` can already support configuring all the things in this PR  in client application already. 

2. Adds a depdency on a crate, it may not be maintained, may not receive security updates (which is already the case with `dotenv`), its maintained fork seems to be `dotenvy`, even then adding a dependency would make sense if it adds substaintial capability/feature. 

3. Introducing non standared new env vars (for base url etc. as in this PR) is also not desirable when people dont expect it in other language libraries, for example official python library. The library already provide option to configure them, where that configuration comes from  - which env var, which file etc. are all flexible options for client application to choose. 

Please let me know if you have additional thoughts on this.

PR #60, opened 2023-03-18
feat: Add rustls support

Xuanwo:
This pull request adds support for rustls, allowing users to switch between options.


64bit:
Hi @Xuanwo , Thank you for the PR to add rustls support. 

It seems that this flips the default behaviour of `reqwest` by using `rustls` instead of `native-tls`, so doing a minor release would be appropriate.

64bit (approved):

64bit:
This is released in v0.10.0

Xuanwo:
> This is released in v0.10.0

Great!

PR #71, opened 2023-05-22
Add completions-proxy example

tot0:
Demonstrates listening for `completions` and `chat/completions` requests on localhost, forwarding them to openai, then printing streamed responses before passing back to local client.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "create-edit",
     "completions",
     "completions-stream",
+    "completions-proxy",
     "create-image",
     "create-image-b64-json",
     "create-image-edit",
--- a/examples/completions-proxy/Cargo.toml
+++ b/examples/completions-proxy/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "completions-proxy"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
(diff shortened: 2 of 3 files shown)


tot0:
Vague plan is to leverage [local.ai](https://github.com/louisgv/local.ai) for serving local models and have support for remote providers as well, this example was me playing around with what would be involved in forwarding openai streaming requests.

64bit:
Hi @tot0 

Thank you for the PR and interesting example. 

So far the examples in this repo are self contained to demonstrate the use of available OpenAI APIs along with their dedicated READMEs.  

The example in this PR lean towards an application built on top, the direction that you wanna go with your plan suggests to me that  having this example in its own repository might be a better choice?

tot0:
Unsure how soon such a future application will materialize so I just wanted
to push the changes at least and create something that helps me remember I
made them :)

If you don’t think they should live in this repo though that’s reasonable
and I can close this.

64bit:
I'm a bit late here to respond.. apologies. 

I don't have any objections  for having it here, however, in spirit of having this example useful for folks new to Rust and new to this repo - do you mind adding:
1.  a README for this example along with a brief overview of what it does
2.  Add ways to actually call the APIs in this example : like `curl` or Postman. So people can copy paste and run your example. 

Above two would make it self documenting where people can actually run it with least friction.

Issue #72, opened 2023-06-13
Add support for functions


m1guelpf:
See https://openai.com/blog/function-calling-and-other-api-updates

PR #73, opened 2023-06-13
Updated schema to v1.3.0

m1guelpf:
Was working on #72 when I realized that part of the types seem to be auto-generated now, but the generating script is nowhere to be found in the repo (at least I couldn't find it).

Could you provide more details about how the process works?

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2,19 +2,19 @@ openapi: 3.0.0
 info:
   title: OpenAI API
   description: APIs for sampling from and fine-tuning language models
-  version: '1.2.0'
+  version: "1.3.0"
 servers:
   - url: https://api.openai.com/v1
 tags:
-- name: OpenAI
-  description: The OpenAI REST API
+  - name: OpenAI
+    description: The OpenAI REST API
 paths:
   /engines:
     get:
       operationId: listEngines
       deprecated: true
       tags:
-      - OpenAI
+        - OpenAI
       summary: Lists the currently available (non-finetuned) models, and provides basic information about each one such as the owner and availability.
       responses:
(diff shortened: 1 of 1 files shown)


64bit:
Hi @m1guelpf ,

Surprised to learn the new announcement from you, great to see you're on top of latest and greatest API updates :) 

Actually this repo it not auto generated, back in Oct/Nov last year I couldn't find generators good enough. Also sometimes the spec is drifted from reality (human error / spec is not updated).  Some of those reasons led me write this by hand - and so in the process I kept the names from spec as it is in code in `types.rs`. 

Given that, updates for fuction calling support are most welcome. I'm happy to merge this based on if you're planning to update this PR or make a separate PR?

m1guelpf:
I'm good to make a separate PR, so feel free to do whatever you think makes the most sense.

As an aside, I had a bunch of Q's regarding implementing functions:

- Do we want to throw a runtime error if you try to call an older model with functions, or just bubble up OpenAI's?

- The model name is currently a String, would it make sense to turn this into an Enum (with an `Other(String)` key ofc). This should make the model check more robust if we go for that.

- I'm planning to pull the jsonschema crate to type the function parameter, would you rather this is behind a feature or available for all?

64bit (approved):
🎉

64bit:
Hey @m1guelpf, I'm ready to ship Azure support in #67 . I wanted to check in with you and see if I should go ahead or wait for any work in progress that you might have based on `main` branch?

64bit:
Shipping #67, hopefully it doesn't affect much, if it does (sorry in advance) but I'm here to help navigate any related issues.

PR #77, opened 2023-06-19
Add more fileds for Model

DaviRain-Su:


async-openai/src/fine_tune.rs
```diff
@@ -76,4 +76,14 @@ impl<'c, C: Config> FineTunes<'c, C> {
             )
             .await)
     }
+
+    /// Delete a fine-tuned model. You must have the Owner role in your organization.
+    pub async fn delete_fine_tun_model(
```
64bit, line 81: This is actually present in `Models::delete` because of the REST endpoint being `/models/<something>` 

https://github.com/64bit/async-openai/blob/main/async-openai/src/model.rs#L33-L37

async-openai/src/types/types.rs
```diff
@@ -12,6 +12,24 @@ pub struct Model {
...
+    pub object: String,
+    pub created: u32,
+    pub allow_create_engine: bool,
+    pub allow_sampling: bool,
+    pub allow_search_indices: bool,
+    pub allow_view: bool,
+    pub allow_fine_tuning: bool,
+    pub organization: String,
+    pub group: Option<String>,
+    pub is_blocking: bool,
```
64bit, line 32: What's this?

DaviRain-Su: according Response Maybe Add it.

64bit:
Can you please provide more context on this PR in the description please?

DaviRain-Su:
just call `openai.Model.list()` response 
```
    {
      "id": "gpt-3.5-turbo-16k",
      "object": "model",
      "created": 1683758102,
      "owned_by": "openai-internal",
      "permission": [
        {
          "id": "modelperm-LMK1z45vFJF9tVUvKb3pZfMG",
          "object": "model_permission",
          "created": 1686799823,
          "allow_create_engine": false,
          "allow_sampling": true,
          "allow_logprobs": true,
          "allow_search_indices": false,
          "allow_view": true,
          "allow_fine_tuning": false,
          "organization": "*",
          "group": null,
          "is_blocking": false
        }
      ],
      "root": "gpt-3.5-turbo-16k",
      "parent": null
    }
```

async-openai/src/fine_tune.rs
DaviRain-Su: fixed, remove it

Issue #76, opened 2023-06-17
Introduce Function Calling, ChatCompletion


Doordashcon:
pls see [documentation](https://platform.openai.com/docs/guides/gpt/function-calling)

64bit:
Duplicate of #72

64bit:
This is available in v0.12.0

PR #79, opened 2023-06-19
Add function call request param

DaviRain-Su:
close: #72 
#76 

- Remove example/Cargo.toml, add to root Cargo.toml
- Add function-call example get_current_weather


b0xtch (approved):

b0xtch:
I would also update the readme to add function as feature

DaviRain-Su:
I am currently writing an example code for function calls, but I encountered an issue with the error: `JSONDeserialize(Error("invalid type: null, expected a string", line: 11, column: 23)).`  I need some help to resolve this.

64bit:
Hi @DaviRain-Su, Thank you for the PR and adding example for this new feature - its a great way to test it.

Here are some hints for you to debug this further:
- see the updated upstream OpenAPI spec and match it with Rust type in code
- setup tracing_subscriber so that `tracing::error!( ... )` can log the actualy json payload causing deserializtion error. `examples/rate-limit-completions` has an example - however you might need to change log level. 

Hope this helps

DaviRain-Su:
Successfully added an example that can use function calls.

async-openai/src/types/types.rs
```diff
@@ -698,17 +699,39 @@ pub enum Role {
 pub struct ChatCompletionRequestMessage {
     /// The role of the author of this message.
     pub role: Role,
+    #[serde(skip_serializing_if = "Option::is_none")]
     /// The contents of the message
-    pub content: String,
+    pub content: Option<String>,
```
64bit, line 704: Comment needs to be updated (please see latest spec)

> The contents of the message. `content` is required for all messages except assistant messages with function calls.

64bit (changes requested):
It appears that naming does not adheres to OpenAPI spec.

For example  `ChatCompletionFunctions` , Please see the names of new types in this PR https://github.com/64bit/async-openai/pull/73


For background: all types in `types.rs` are named based on OpenAPI component schema.

DaviRain-Su:
Have fix all.

64bit:
`function_call` in `ChatCompletionRequestMessage` and `ChatCompletionResponseMessage` can also be a struct with `name` and `arguments` - I'll take care of it before publishing.

Thank you for addressing the comments,  nice work on the new example!

64bit (approved):

64bit:
This is released in v0.12.0

PR #82, opened 2023-06-28
Add reqwest/native-tls-vendored feature

mikeparisstuff:
Hey, first of all thanks for the awesome crate.

I run my rust services on ARM-based AWS graviton instances which have a different version of OpenSSL that doesn't play nicely with reqwest. This diff adds a new feature to async-openai that allows removing the dependency on OpenSSL by opting in for the native-tls-vendored feature in the reqwest library. 

I've tested this on graviton with an incompatible version of OpenSSL and things work nicely :)

Thanks


64bit (approved):
Thank you for PR and detailed description & testing, nicely done!

64bit:
This is released in v0.12.1

PR #83, opened 2023-07-01
Fix streaming function calls

crat0z:
Hello!

This fixes a bug that broke streaming chat responses with function calls. From observation, `arguments` is always a string, however, just to be safe, I've made it an `Option<String>`.


64bit:
Hi @crat0z , thank you for the PR and fixing the bug. 

Do you mind adding the full example in  (say) `examples/function-call-stream` if you already have it from your testing? it would be very helpful to test & maintain this in the future as well

64bit:
Thank you for PR @crat0z , this was addressed in #98

PR #88, opened 2023-07-25
Fix api base in url

kirillkuzin:
Hello! Thanks for this crate, which I'm using in my projects.
I found the bug: even if you specify api base in the config, the client will replace it when making a request (for openai config). So, this is the fix :)

### DIFF
--- a/async-openai/src/config.rs
+++ b/async-openai/src/config.rs
@@ -84,7 +84,7 @@ impl Config for OpenAIConfig {
     }
 
     fn url(&self, path: &str) -> String {
-        format!("{}{}", OPENAI_API_BASE, path)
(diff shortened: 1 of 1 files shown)


64bit (approved):
Hi @kirillkuzin 

Always a pleasure to hear from library users :) 

Good find and thank you for fixing the bug!

64bit:
Even though it still is bug in library for correctness,  I recalled why it was like that. 

I made an assumption that API endpoint for OpenAI will highly likely not change - is that still a good assumption?

64bit:
This is released in v0.12.2

kirillkuzin:
Actually, I don't know the precise cases for this. I have received a message about that from one guy using my plugin.

Issue #87, opened 2023-07-21
Complimentary crate for async-openai: "openai-func-enums"


frankfralick:
async-openai is 😙👌, but I found myself with a lot of function-call related logic to deal with, and much more still to come, and inlining json objects isn't why I came to the Rust party.  [openai-func-enums](https://github.com/frankfralick/openai-func-enums) lets you compose descriptions of "functions" using enum types and enum variants to represent functions and their arguments, generate the function json, and deserialize into struct instances (types generated by the library) that have properties matching your enum field types, so you can match on the variants and get on with it. Just thought I'd share in case it was useful to anyone else.

64bit:
Hi @frankfralick 

openai-func-enums is a great idea, love your approach! Thank you for sharing. 
For people to discover it easily - You're welcome to link it as complimentary crate or related crates in main README or/and examples/function-call/README.

64bit:
Resolved by #89

64bit:
Hello, 
I have removed this from readme - because of so many breaking changes in DevDay release. Please do send PR if the openai-func-enums is updated. 

Thank you

frankfralick:
Thanks for the heads up!

frankfralick:
The ability to have it call functions in parallel is so crazy in terms of what it makes possible. I'm pretty much done updating things to work with version 16.2.  Just wanting to first check to see if you think 16.2 is fairly stable at this point in terms of tools and tool calls vs. functions/function calls? 

Also, I see Issue #143 asking for an example using tools instead of deprecated function_calls, etc. Are you wanting two new examples for this like tool-call and tool-call-stream just dealing with chat completions and leave the examples dealing with the deprecated fields the way they are? I'm happy to do them. 

Currently I have it where it runs parallel calls that openai returns using tokio::spawn, but I think I'
![image](https://github.com/64bit/async-openai/assets/2084232/cc453c16-ba8d-4c75-9852-d9edf11d2398)

64bit:
Great to hear that you have it updated, will be happy to merge PR for companion crate again :) 

Even though I have done all due diligence in shipping things are always prone to human errors - either in upstream spec or my own. Specially when this was a really big change (please see size of #131 ) and its all hand written without any code generation. 

Please excuse if you find rough edges. 

With that out of the way, two new examples using tools are most welcome!

PR #89, opened 2023-07-27
Link to openai-func-enums

frankfralick:
This pull request is in response to issue #87.  Also, since I opened this, @WestXu asked about integration with [clap](https://github.com/clap-rs/clap), which I've done now and it's pretty fun. You can see an example of integrating async-openai with clap [here](https://github.com/frankfralick/openai-func-enums#integration-with-clap). Thanks again.


64bit (approved):
Clap example looks like magic! 
Thank you for the PR!

Issue #85, opened 2023-07-20
Add example for Function Call with streaming 


64bit:
Just like `examples/function-call` but using streaming call. 

Note: There is open PR at the moment but no way to test it #83

victor-wei126:
@64bit going to try to tackle this issue first.

I've been skimming the codebase and also the OpenAPI spec. Just curious (and this might help me understand the codebase better) - what was the main design and considerations you took while developing this library to use the OpenAI API? Did you mostly use the OpenAPI spec or look at the design of their official python library?

64bit:
The primary source was OpenAPI spec and async-stripe was source of inspiration, you might find evolution of this library if you look at early commits. When I started I couldnt find good resources to generate code from spec, and json spec is not always up to date, sometimes has human errors. Moreover there are complex situations like streaming where an input field like `"stream": true` determines SSE response. 

So I decided to build this library with hand instead of relying on auto generation (although auto generation would be ideal for Rust in long term) because my focus was on correctness, auto retries on rate limit and full API coverage - all the things to make library production ready.

I had skimmed through the offcial python library too, and then went with spec as source of truth and grouping ("images", "fine-tunes" etc.) based on OpenAPI spec + RESTful spec.

64bit:
resolved  in #90

PR #90, opened 2023-07-29
add function-call-stream example

victor-wei126:
This addresses issue #85.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "create-image-variation",
     "embeddings",
     "function-call",
+    "function-call-stream",
     "models",
     "moderations",
     "rate-limit-completions",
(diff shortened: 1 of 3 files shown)


victor-wei126:
I've looked at PR #83. This PR does not take those changes into account.

I'm not completely sure what that PR does, but from looking at it I suppose that PR allows streaming the "first" model response, ie, the function name and arguments that the model has generated in order for the function to then be called with those arguments?

64bit:
Nice work, thank you @victor-wei126 for PR! 

Perhaps PR 83 assumes that streaming doesn't contain all the fields in response (for example chat streaming has different response type then its non-streaming counterpart) and replaces them with `Option<T>`. 

Because this example works, seems like we don't need to consider changes from PR 83. 

---- 

Regarding PR: I recommend running `cargo fmt` for standard formatting for future.

64bit (approved):

Issue #75, opened 2023-06-16
Use official types for chat completion stream response from spec


64bit:
`async-openai`'s existing types for Chat stream were written from observation of OpenAI response. 

Recently (16th June 2023) officially types were update for it, and needs to be used in `async-openai` as well 
Ref: https://github.com/openai/openai-openapi/pull/48/files

`async-openai` has used the type names from spec for majority of types  - hence `ChatCompletionStreamResponseDelta` and `CreateChatCompletionStreamResponse` names should be used.

0xVesion:
Hey, I'd be happy to pick this one up. Please give me some pointers on how to proceed. Should I generate the types using the OpenAPI spec and replace the existing ones in src/types/types.rs?

64bit:
Hi Vincent,

Thank you for your interest and offering to help! Actually `async-openai`
is written by hand and is not generated for various reasons (human error in
spec - this issue is an example of such, lack of tooling at the time I
created it etc.)

For resolving this issue, it requires the following:
1 See the function signature in `chat.rs` for `create_stream` its response
type would change to the ones in
https://github.com/openai/openai-openapi/pull/48/files
2. Hence corresponding types need to be updated in `types.rs`
3. Example need to be updated with new types `examples/chat-stream`

Let me know if you need further clarification!

Thank you,
Himanshu

Doordashcon:
Hello @64bit, looks like `CreateChatCompletionStreamResponse` up to date [here](https://github.com/64bit/async-openai/blob/b4e36716df4ba09446ac42fc1e303f6187d55865/async-openai/src/types/types.rs#L881C46-L881C46)

64bit:
Hi @Doordashcon , Ah pure coincidece however its field `choices`  with vector of  `ChatChoiceDelta` should be vector of `ChatCompletionStreamResponseDelta`.  ( There could be more - but I haven't parity checked with spec for this comment)

victor-wei126:
I'm interested in contributing a solution to this issue. Is this project still active?

64bit:
Hi @victor-wei126 

Thank you for offering to contribute ❤️ 

Indeed the project is active, there was a new version release a few weeks ago. 

For this particular issue, I'd recommend to have latest spec from upstream repo, so you can see the diff of updated spec with the one in this repo and would help to update types as well.

victor-wei126:
Hey @64bit ... my pleasure :) - I'm interested in AI and Rust - hope I can be helpful.

[...]

PR #92, opened 2023-08-01
[WIP] modify types for chat completion stream response

oslfmt:
Attempt to solve issue #75.

I've changed the following types to match the OpenAI API spec. Couple of questions:
- Which types do you determine to be "Option"? I've stripped the option type from any required types as specified by the API spec, but not sure if option should be applied to the other types or not
- I've noticed there's an enum type `ChatCompletionFunctionCall` and just a `FunctionCall` type. Not sure which I should use.

Beyond that, how do you go about updating the yaml file? As the one in the current repo is out of date. Should I pull the updated spec into this repo as well?


64bit:
Thank you @victor-wei126. 

> Which types do you determine to be "Option"? I've stripped the option type from any required types as specified by the API spec, but not sure if option should be applied to the other types or not
I determinte `Option` from the spec, the fields which are not in the `required` list are `Option`als. 

> I've noticed there's an enum type ChatCompletionFunctionCall and just a FunctionCall type. Not sure which I should use.
The best is to look at OpenAPI spec to determine which one (sometimes spec doesnt have details about streaming responses - then the best is to execute and see API response)

> Beyond that, how do you go about updating the yaml file? As the one in the current repo is out of date. Should I pull the updated spec into this repo as well?
My usual practice was to get upstream spec in my pull request AND the diff was what I used to update the rest of the code / types.rs

victor-wei126:
@64bit Thanks for the response.

I've updated the types now to reflect what I think is proper (Option or not), trying to reflect the latest spec as closely as possible. The main question is whether to leave ID, object, and created types as option or not, as @lpi mentioned it is easier as openrouter.ai does not provide them.

> The best is to look at OpenAPI spec to determine which one
`FunctionCall` and `ChatCompletionFunctionCall` are types define in async-openai library, not the official specification itself. I choose to use `FunctionCall` type as this more closely resembles the official spec.

> My usual practice was to get upstream spec in my pull request AND the diff was what I used to update the rest of the code / types.rs
Perfect, I'll do this then. I'll also test against the API response as you mentioned above. If I have any questions, will let you know.

Just out of curiosity - is there any way to automatically test the types against the yaml spec? I read a bit about OpenAPI and saw they have some functionality regarding automatic code generation. Not entirely sure about it though - was wondering if you had used this feature so I can speed up the development here if possible.

64bit:
> By upstream spec do you mean the current spec in this repo or the latest openai-openapi spec?

I mean openai-openapi spec, the source of truth.

> Just out of curiosity - is there any way to automatically test the types against the yaml spec?

I shared my experience with auto generation in this comment https://github.com/64bit/async-openai/issues/85#issuecomment-1646752463 

I think for auto generation to be practical, there's a need for a good openapi spec to Rust generator, with ability to customize generated names. (There could be one - I haven't searched in a long time) 

But for now the best way to test is using examples. They serve two purposes: documentation for users and also testing when making changes.

oslfmt:
@64bit this PR is ready for review

64bit:
Thanks!

I'll take a look, btw seems like your Github account changed?

64bit (approved):

64bit:
This is released in v0.13.0

oslfmt:
Yup haha changed my name but still me

Issue #94, opened 2023-08-05
Azure OpenAI Errors are unhandled


xangelix:
Here's one example that only gives a JSONDeserialize error.

```json
{
  "error": {
    "code": 429,
    "message":  "Requests to the Creates a completion for the chat message Operation under Azure OpenAI API version 2023-05-15 have exceeded token rate limit of your current OpenAI S0 pricing tier. Please retry after 60 seconds. Please go here: https://aka.ms/oai/quotaincrease if you would like to further increase the default rate limit."
  }
}
```

JSONDeserialize(Error("missing field `type`"))

64bit:
Thank you @xangelix for reporting the issue. 

This is because of unfortunate differences in specs for Error:
-   https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2023-03-15-preview/inference.json#L774C25-L796
- https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L1579-L1598

Quick fix could be making `r#type` field in `ApiError` as `Option<String>`.

ChristopherJMiller:
This is affecting me as well, I'll work on at least implementing the quick fix, and come back to it if more should be done 😃

Ideally from an end-library-user perspective I want the error visible so I know what's wrong

64bit:
Resolved by #95

PR #95, opened 2023-08-08
Made ApiError `r#type` optional to support Azure Error responses

ChristopherJMiller:
Made `ApiError::r#type` an optional string to support the mismatch between Azure and OpenAI's specification. This will at least let Azure users view the actual error instead of receiving JSON Deserialization errors.

Discussed in #94

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -231,7 +231,7 @@ impl<C: Config> Client<C> {
                         if status.as_u16() == 429
                             // API returns 429 also when:
                             // "You exceeded your current quota, please check your plan and billing details."
-                            && wrapped_error.error.r#type != "insufficient_quota"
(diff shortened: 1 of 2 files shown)


64bit:
Thank you @ChristopherJMiller  for the PR. 

Based purely on spec it seems `message` would also need to be `Option`al?

But if this works without making `message` optional, we can ship this.

ChristopherJMiller:
In local testing with my azure resource I'm now getting the full `ApiError` error results instead of `JSONDeserialize` error results, I think we're good to go! 😃

I'll keep an eye out if I encounter message also being not sent and report back

64bit (approved):
Thank you for confirming!

64bit:
This is released in v0.13.0

PR #98, opened 2023-08-17
Made function call streamable

buraktabn:
Making `FunctionCall`'s arguments `Optional` allow us to stream the functions. Otherwise, it throws a parsing error.


64bit:
Hi @buraktabn ,

Thank you for PR! This is a really good update (and bug fix) to make example fully streamable. 

To maintain parity with OpenAPI spec this PR requires following changes: 

1. Keep existing `FunctionCall` as it is because it represents non-streaming response : https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L2124-L2136

2. Introduce, lets say, `FunctionCallStream` with the `Option<String>` to represent streaming response: https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L2211-L2220

3. Use `FunctionCallStream` to be the type for `function_call` field for `ChatCompletionStreamResponseDelta` struct.

buraktabn:
Hi @64bit,

Thanks for the review. I've made the changes as you suggested and pushed the commit Let me know if anything else is needed.

64bit:
This PR is released in v0.13.1

PR #100, opened 2023-08-23
Retry form submissions on rate limit

### LINKED ISSUE #86: Make Form submission retryable on rate limiting
form submissions are currently not retried when rate limited because the `Form` object used is not clonable. 

Ideas: May be use clonable form object in "group" to be passed down to client.rs and then create "form" object that `reqwest` needs. Going from cloneable to uncloneable.

64bit:
Closes #86 

- all API are now retried on rate limiit (except SSE streaming ones)
- update rate limit example (even though edit API as been deprecated - its has the lowest rate limit and easy to test)
- update readme/docs


stevenhuyn:
Does this mean for this + 14.0 release, `rate-limit-edit` is more for overriding the default rate limit?

64bit:
No, the example makes many api calls back to back triggering actual rate limit on OpenAI so that retries can be tested.

PR #108, opened 2023-09-14
add  deserialization for OpenAIConfig and AzureConfig

### LINKED ISSUE #107: Add `Deserialize` to `OpenAIConfig` and `AzureConfig`
Please add `Deserialize` derive to `OpenAIConfig` and `AzureConfig`. I store my Azure configs in a json file for testing. Without `Deserialize`, I need to copy a same struct and add `Deserialize` for it and convert it to `AzureConfig`, which is unnecessarily verbose.

ifsheldon:
This should close #107. Arguably, the deserialization uses `Default` when a field is not given during deserialization and thus is too permissive. As far as I know,  Azure APIs needs all fields in `AzureConfig` to be valid ones, but the defaults are not.


64bit (approved):
Thank you for the PR!

64bit:
This is released in v0.14.2

Issue #74, opened 2023-06-16
Azure OpenAI Service: Completions Stream ends in "Stream failed: stream ended"


64bit:
See https://github.com/64bit/async-openai/pull/67#issuecomment-1554031714

PR #110, opened 2023-09-16
Support Azure Streaming

dougtallmadge:
Addresses https://github.com/64bit/async-openai/issues/74 and adds support for Embedding Deserialization. This also allows for streaming completion to fail gracefully on any Stream ended event as supposed to hanging until a timeout.

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -302,6 +302,10 @@ where
         while let Some(ev) = event_source.next().await {
             match ev {
                 Err(e) => {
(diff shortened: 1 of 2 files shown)


64bit:
Thanks for the fix and PR @dougtallmadge 

Interesting fix, I'm still left to wonder why `[DONE]` marker doesn't work as expected. 

There's a corresponding test (commented out) in `examples/azure-openai-service` to help test this -  it would be much appreciated if you can see that works with this fix?

Issue #93, opened 2023-08-03
Update library to match OpenAI API release 2.0.0


oslfmt:
OpenAI released a new version of their API [here](https://github.com/openai/openai-openapi/releases/tag/2.0.0).

Major changes are removing deprecated endpoints as well as updating the ```CreateChatCompletionStreamResponse``` and creating ```ChatCompletionStreamResponseDelta```. Those types are brought up in #75.

64bit:
This is released in v0.15.0

PR #111, opened 2023-09-24
chore: Add enum type for finish_reason

retrage:
The `finish_reason` is defined as an enum type in the OpenAI API spec: https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L2821.


64bit:
Hi @retrage 

Thank you for  PR updating the type with spec. 

Since this introduces breaking change, and its unknown to me how much spec has been updated - so I'll merge this PR when I get to sync code with full latest spec to avoid released multiple breaking changes. 

Please see #93

64bit (approved):
Thank you for PR, its ready to be shipped!

PR #112, opened 2023-09-24
do not print key in Debug

Emilgardis:


64bit:
Hi @Emilgardis Thank you for the PR! 

Actually `AzureConfig` suffers from the same security issue. Might as well fix both in one go?

Emilgardis:
Done!

64bit (approved):
Thank you!

64bit:
This is released in v0.14.3

TmLev:
It also makes sense to use https://docs.rs/secrecy/latest/secrecy/

64bit:
Thanks @TmLev thats actually a good suggestion. 
I'm tracking it here https://github.com/64bit/async-openai/issues/113

Issue #113, opened 2023-09-25
security: Use `secrecy` crate to store API key


64bit:
In `OpenAIConfig` and `AzureConfig` handle API key via https://docs.rs/secrecy/latest/secrecy/

64bit:
This is released in v0.15.0

PR #114, opened 2023-09-26
Use `secrecy` for OpenAIConfig and AzureConfig

ironman5366:
Addresses #113 

`println!("{:?}", config)`:

`OpenAIConfig { api_base: "https://api.openai.com/v1", api_key: Secret([REDACTED alloc::string::String]), org_id: "" }`

You have a great project here!

### DIFF
--- a/async-openai/src/config.rs
+++ b/async-openai/src/config.rs
@@ -1,5 +1,6 @@
 //! Client configurations: [OpenAIConfig] for OpenAI, [AzureConfig] for Azure OpenAI Service.
 use reqwest::header::{HeaderMap, AUTHORIZATION};
+use secrecy::{ExposeSecret, Secret};
 use serde::Deserialize;
 
 /// Default v1 API base url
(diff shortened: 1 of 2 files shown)


async-openai/src/config.rs
```diff
@@ -104,7 +95,7 @@ impl Config for OpenAIConfig {
     }
 
     fn api_key(&self) -> &str {
-        &self.api_key
+        &self.api_key.expose_secret()
```
64bit, line 98: This actually leaks the protection proved by `Secret<String>` in first place. 

This calls for a change to function signature to  `fn api_key() -> &Secret<String>` in `Config` trait and in both implementations.

64bit:
Hi @ironman5366 

Thank you for the PR, it needs a small change where we don't expose secrets without `Secret<String>` to clients. That way benefit of it propogates to all applications built using async-openai

async-openai/src/config.rs
ironman5366: Ah, ofc, https://github.com/64bit/async-openai/pull/114/commits/f5dade574e42d7770f9355081aa78023873c2849

64bit (approved):
Thank you. 

I'll bundle it in next release with other breaking changes, probably v0.15.0.

TmLev:
It would be nice to re-export `secrecy::ExposeSecret` trait so that downstream users don't need to add `secrecy` crate to their applications just for that

64bit:
Dang! It doesn't work without adding dependency?

PR #117, opened 2023-10-10
Derive Copy for enums

m1guelpf:
Enums that do not store additional data can be copied around.


64bit (approved):
Nice, thank you for PR!

PR #118, opened 2023-10-10
Better ChatCompletionFunctionCall interface

m1guelpf:
According to [OpenAI's docs](https://platform.openai.com/docs/api-reference/chat/create), the `function_call` parameter can only be either `none`, `auto`, or `{ "name": function_name }`.

The current interface in this library feels counter-intuitive, and I've often tried to call a custom function using `ChatCompletionFunctionCall::String(function_name)` instead of `ChatCompletionFunctionCall::Object(json!({ "name": custom_function }))`.

This change proposes a more intuitive interface for this (as a breaking change unfortunately).

### DIFF
--- a/async-openai/src/types/impls.rs
+++ b/async-openai/src/types/impls.rs
@@ -356,7 +356,7 @@ impl From<&str> for ChatCompletionFunctionCall {
         match value {
             "none" => Self::None,
             "auto" => Self::Auto,
-            _ => Self::Function(value.to_string()),
+            _ => Self::Function { name: value.to_string() },
         }
     }
 }
--- a/async-openai/src/types/impls.rs
+++ b/async-openai/src/types/impls.rs
@@ -351,13 +351,36 @@ impl_from_for_array_of_integer_array!(u16, Prompt);
 
 impl From<&str> for ChatCompletionFunctionCall {
     fn from(value: &str) -> Self {
-        ChatCompletionFunctionCall::String(value.to_string())
+        match value {
+            "none" => Self::None,
+            "auto" => Self::Auto,
+            _ => Self::Function(value.to_string()),
+        }
     }
 }
 
-impl From<serde_json::Value> for ChatCompletionFunctionCall {
-    fn from(value: serde_json::Value) -> Self {
-        ChatCompletionFunctionCall::Object(value)
+impl Serialize for ChatCompletionFunctionCall {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        match self {
+            ChatCompletionFunctionCall::None => serializer.serialize_str("none"),
+            ChatCompletionFunctionCall::Auto => serializer.serialize_str("auto"),
+            ChatCompletionFunctionCall::Function(s) => {
+                let mut map = serializer.serialize_map(Some(1))?;
+                map.serialize_entry("name", s)?;
+                map.end()
+            }
+        }
+    }
+}
+
+impl<'de> serde::de::Deserialize<'de> for ChatCompletionFunctionCall {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+
+        Ok(s.as_str().into())
     }
 }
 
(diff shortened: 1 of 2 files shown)


64bit:
Actually previous `From` [conversion trait implemented](https://github.com/64bit/async-openai/blob/main/async-openai/src/types/impls.rs#L352-L362) lets you provide function_call parameter without even using `ChatCompletionFunctionCall`:

```
CreateChatCompletionRequestArgs::default()
    .function_call("auto") //  or "none" 
```
OR

```
CreateChatCompletionRequestArgs::default()
    .function_call(json!({ "name" : "hello_world" })
```

--- 

But yes `enum` does offer better API. 

No worries about breaking changes, because next release v0.15.0 will have other breaking change (already in `main`) and possible breaking updates from latest openapi spec.  

I'll merge this PR once spec sync is addressed too for v0.15.0.

async-openai/src/types/impls.rs
```diff
@@ -351,13 +353,36 @@ impl_from_for_array_of_integer_array!(u16, Prompt);
...
+}
+
+impl<'de> serde::de::Deserialize<'de> for ChatCompletionFunctionCall {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+
+        Ok(s.as_str().into())
```
64bit, line 385:
```suggestion

```
Custom serialization implementation is not required this can be handled by Serde

async-openai/src/types/types.rs
```diff
@@ -37,11 +37,14 @@ pub enum Stop {
...
+#[derive(Debug, Clone, PartialEq)]
 pub enum ChatCompletionFunctionCall {
-    String(String),
-    Object(serde_json::Value),
+    /// The model does not call a function, and responds to the end-user.
+    None,
+    /// The model can pick between an end-user or calling a function.
+    Auto,
+    /// Forces the model to call the specified function.
+    Function(String),
```
64bit, line 47:
```suggestion
    Function { name : String },
```

async-openai/src/types/impls.rs
```diff
@@ -351,13 +353,14 @@ impl_from_for_array_of_integer_array!(u16, Prompt);
 
 impl From<&str> for ChatCompletionFunctionCall {
     fn from(value: &str) -> Self {
-        ChatCompletionFunctionCall::String(value.to_string())
+        match value {
+            "none" => Self::None,
+            "auto" => Self::Auto,
+            _ => Self::Function(value.to_string()),
```
64bit, line 359:
```suggestion
            _ => Self::Function { name: value.to_string() },
```

async-openai/src/types/types.rs
```diff
@@ -781,10 +784,7 @@ pub struct CreateChatCompletionRequest {
     pub functions: Option<Vec<ChatCompletionFunctions>>,
 
     /// Controls how the model responds to function calls.
-    /// "none" means the model does not call a function, and responds to the end-user.
-    /// "auto" means the model can pick between an end-user or calling a function.
-    /// Specifying a particular function via `{"name":\ "my_function"}` forces the model to call that function.
-    /// "none" is the default when no functions are present. "auto" is the default if functions are present.
+    /// None is the default when no functions are present. Auto is the default if functions are present.
```
64bit, line 787:
```suggestion
    /// "none" means the model does not call a function, and responds to the end-user.
    /// "auto" means the model can pick between an end-user or calling a function.
    /// Specifying a particular function via `{"name":\ "my_function"}` forces the model to call that function.
    /// "none" is the default when no functions are present. "auto" is the default if functions are present.
```

Keep comment for docs.rs

async-openai/src/types/impls.rs
```diff
@@ -3,6 +3,8 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use serde::{ser::SerializeMap, Serialize};
+
```
64bit, line 7:
```suggestion
```
Not required anymore

64bit:
Hello @m1guelpf ,

I took liberty to make some changes and kept your idea intact. Essentially  use serde instead custom serialization implementation.

Thank you for your patience on this PR!

64bit (approved):

PR #120, opened 2023-10-15
add feature gate for tokio

### LINKED ISSUE #102: Wasm support
Thanks for making this crate, it seems very useful :)

Currently the crate always depends on e.g. tokio which means it can't be compiled to wasm for use in frontends (or serverless wasm workers like on AWS/Cloudflare) that want to make OpenAI API requests.
It would be great if this crate could also be compiled to wasm.

ifsheldon:
This adds a feature gate for all functionalities that requires `tokio`, and it should pave the way to close #102. Adding the feature gate `enable_tokio` should not be a breaking change.

Please note that since I don't have an API key from OpenAI but from Azure OpenAI, I couldn't test if these changes break OpenAI functionalities, so please help me test them. Thanks!

### DIFF
--- a/async-openai/Cargo.toml
+++ b/async-openai/Cargo.toml
@@ -15,7 +15,9 @@ homepage = "https://github.com/64bit/async-openai"
 repository = "https://github.com/64bit/async-openai"
 
 [features]
-default = ["rustls"]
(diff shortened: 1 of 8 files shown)


ifsheldon:
closing this as it's merged into #122

PR #123, opened 2023-10-17
Expose `Config`

Czechh:
A good practice in production environments is to have multiple clients from multiple endpoints and api keys to increase reliability and throughput. But, managing these different clients can be hectic as some fail. Providing a way to debug more easily by exposing the config would be great. A workaround would be to wrap every client in a container with the config info, but it's not ideal.

This simply exposes `Config` parameter so we can do something like `client.config().api_base()`

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -110,6 +110,10 @@ impl<C: Config> Client<C> {
         Audio::new(self)
     }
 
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you for context and PR! 

Please consider this to be part of next v0.15.0 release sometime this month.

PR #130, opened 2023-11-06
Add speech endpoint

m1guelpf:
- Added a `post_raw` and `execute_raw` internal functions, since we don't want to JSON-decode the response body for the speech endpoints (which return raw audio). Modified the `execute` function to wrap around `execute_raw`.
- Added basic support for the TTS api

Note that the struct corresponding to the `response_format` parameter was named `SpeechResponseFormat` instead of `AudioResponseFormat`, since the later already existed for defining whisper response formats.

I'd love to add an enum for voices as well (with an Other wildcard and marked as non-exhaustive), but refrained since the library's pattern seems to be deferring to strings (same case as model IDs all across).

### DIFF
--- a/async-openai/Cargo.toml
+++ b/async-openai/Cargo.toml
@@ -40,6 +40,7 @@ tracing = "0.1.37"
 derive_builder = "0.12.0"
 async-convert = "1.0.0"
 secrecy = { version = "0.8.0", features=["serde"] }
+bytes = "1.5.0"
 
 [dev-dependencies]
 tokio-test = "0.4.2"
--- a/async-openai/README.md
+++ b/async-openai/README.md
@@ -23,7 +23,7 @@
 
 - It's based on [OpenAI OpenAPI spec](https://github.com/openai/openai-openapi)
 - Current features:
-  - [x] Audio
+  - [x] Audio (Whisper/TTS)
   - [x] Chat
   - [x] Completions (Legacy)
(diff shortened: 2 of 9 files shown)


64bit:
Thank you! 🎉 

( There's https://crates.io/crates/serde_bytes too, not sure what difference would be between `bytes` create and `serde_bytes` )

> 'd love to add an enum for voices as well (with an Other wildcard and marked as non-exhaustive), but refrained since the library's pattern seems to be deferring to strings (same case as model IDs all across).

I think for Voice as enum makes sense - go for it . I usually defer to spec when it lists a filed of type string + enum. They haven't released updated spec though. 

 ----

I'm getting "The model `tts-1` does not exist or you do not have access to it."  Not sure how to get access to the new models?

m1guelpf:
Not sure what the difference is either, but `bytes` is the type `reqwest` uses, we just import the crate to access the type really.

Will add the enum in a sec. Model releases in 1h (supposedly) so not available yet lmeow

64bit:
Aha here, https://openai.com/blog/new-models-and-developer-products-announced-at-devday

>We’ll begin rolling out new features to OpenAI customers starting at 1pm PT today.

30mins to go 🕐

64bit (approved):
Looking good. will target releasing this in v0.16.0 with other new APIs

PR #141, opened 2023-11-09
Make Run.expires_at optional

nodir-t:
Sometimes Run retreivals fail with message that it couldn't convert null to i32. Looks like expires_at is actually optional, see example in https://platform.openai.com/docs/api-reference/runs

Change its type from i32 to Optional<i32>. The problem was resolved after this change.


64bit:
Oh its one of those cases where spec has a diff. 

Thank you for the description and fix!

PR #147, opened 2023-11-14
Rename tools_outputs to tool_outputs

nodir-t:
Field SubmitToolOutputsRunRequest.tools_outputs has a typo. It should be be called tool_outputs instead.
This PR renames it to tool_outputs and fixes deserialization.


64bit:
Thank you for the fix!

Issue #151, opened 2023-11-20
Example of multiple function calls


ysimonson:
GPT-4 Turbo supports multiple function calls, but I couldn't any examples of this in async-openai yet. It took me a little while to stitch together how to do this since it's a bit divergent from single function calls, so maybe this minimal demonstration will be helpful to someone else:

```rust
let request = CreateChatCompletionRequestArgs::default()
    .model("gpt-4-1106-preview")
    .messages([
        ChatCompletionRequestUserMessageArgs::default()
            .content(prompt)
            .build()?
            .into(),
        ChatCompletionRequestSystemMessageArgs::default()
            .content(SYSTEM_PROMPT)
            .build()?
            .into(),
    ])
    .max_tokens(512u16)
    .stream(false)
    .tools([
        ChatCompletionToolArgs::default()
            .r#type(ChatCompletionToolType::Function)
            .function(
                // ... build ChatCompletionFunctions here ...
            )
            .build()?,
        ...
    ])
    .build()?;

if let Some(tool_calls) = response_message.tool_calls {
    tool_calls
        .into_iter()
        .map(|tool_call| ...)
        .collect()?;
}
```

64bit:
Seems like this was addressed in #153? yes?

ysimonson:
Yup!

Issue #143, opened 2023-11-10
Provide function call example using tool_calls


sloganking:
https://github.com/64bit/async-openai/blob/main/examples/function-call/src/main.rs

The current function call example uses the now deprecated `ChatCompletionResponseMessage::function_call` field. It was deprecated in favor of `ChatCompletionResponseMessage::tool_calls`. Can an example be provided using the new `tool_calls` method?

SpaghettiFibonacci:
Here is an example of how to use it meanwhile:

```rust
let tool = ChatCompletionToolArgs::default()
        .r#type(async_openai::types::ChatCompletionToolType::Function)
        .function(ChatCompletionFunctions { 
                name: "get_current_weather".to_string(), 
                description: Some("Get weather in location".to_string()), 
                parameters: json!({
                    "type": "object",
                    "properties": {
                        "location": {
                            "type": "string",
                            "description": "The city and state, e.g. San Francisco, CA",
                        },
                        "format": {
                            "type": "string",
                            "enum": ["celsius", "fahrenheit"],
                            "description": "The temperature unit to use. Infer this from the users location.",
                        },
                    },
                    "required": ["location", "format"],
                    
                }) 
            }).build()?;
```

Though I am not sure it is working right now!
```
Error: ApiError(ApiError { message: "'$.tool_choice' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", type: Some("invalid_request_error"), param: None, code: None })
```

sloganking:
Okay thanks, I will continue using the deprecated option for now and maybe come back to this when it has been tested further. or test it some on my own later.

SpaghettiFibonacci:
> ```rust
> ```rust
>         .tool_choice(ChatCompletionToolChoiceOption::Auto)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

When leaving out ```
        .tool_choice(ChatCompletionToolChoiceOption::Auto)

``` 

it should work

64bit:
An example was added in #153. And the  bug  was fixed in #158 

Seems like everything is resolved, hence closing the issue.

PR #153, opened 2023-11-21
Add examples tool-call and tool-call-stream

frankfralick:
This PR adds non-streaming and streaming examples of chat completions with tool calls. These two examples are fairly close to the function-call and function-call-stream examples that use the deprecated parts of api. These examples keep the "get_current_weather" function but change the prompt to ask for the weather in two locations. This is to induce multiple tool calls in order to demonstrate dealing with the possibility of parallel tool calls.

This should address issue #143 (provide tool call examples) and issue #151 (provide example handling parallel/multiple tool calls).

### DIFF
--- a/examples/tool-call-stream/Cargo.toml
+++ b/examples/tool-call-stream/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "tool-call-stream"
+version = "0.1.0"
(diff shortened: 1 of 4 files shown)


64bit:
Wow this looks like a lot of work, thank you!

Issue #144, opened 2023-11-10
FIleInput from memory


nodir-t:
I am trying to understand how to upload a file with contents from memory and can't figure out how. It appears that file input is only from a file? Do I have to save my bytes to file system to read them back?

Prosammer:
@64bit I am looking for the ability to add audio data vecs to AudioInput instead of having to write them to a file first, would this be something you'd be interested in a PR for?

64bit:
Absolutely interested!I think having this feature would work same for bunch of inputs: FileInput, ImageInput, AudioInput. Looks like they need to be enum to support paths and bytes?

Regarding PR: for non trivial features like this one it really helps to have a new self contained example because it becomes easy to test new feature and old examples.

nodir-t:
Could you accept `io:Read` instead, not a Vec? I'd like to be able to pipe data from a network request to OpenAI without having it buffer it neither in the file system, nor memory.

Prosammer:
Hey @nodir-t @64bit , I put up a PR which works for in-memory files (`Bytes` and `vec[u8]` as that's what `reqwest::Body` accepts), but I wasn't able to figure out a way to get these inputs to take an `io::Read` as it seems `reqwest::Body` wants `AsyncRead`. I tried having these Inputs accept a generic that was `AsyncRead` but got confused as to how that would work with the derive_builder builder which would require the `AsyncRead` to be `Clone` as well. I'm fairly new to Rust and open source PRs so any feedback would be much appreciated!

64bit:
hey @Prosammer , thank you I was able to check it out today and it looks good given that its backward compatible change for consumers of library happy to merge it.

I'll keep this issue open for the context and if someone wants to continue work on `AsyncRead` and `io:Read` suggested in this thread.

ifsheldon:
This seems coincidentally solve some issues for wasm support. Thanks @Prosammer! If you have some time and interest, you can take a look at the experiment branch and help mature the wasm support.

Prosammer:
@ifsheldon haha awesome, I'm not very familiar with wasm but I will take a look!

nodir-t:
`AsyncRead` is even better. Thank you

PR #154, opened 2023-11-21
In memory files

Prosammer:
As per #144, added ability to pass Bytes or vec[u8]s directly for ImageInputs, FileInputs, and AudioInputs.


64bit:
Thank you this is nice work, and example makes my life easy to test :)

PR #155, opened 2023-11-22
FIX: finetuningjob deserialize crash

inszee:
let job = open_ai_client
            .fine_tuning()
            .create(fine_tune_request)
            .await?;
FineTuningJob struct is like:  
{
  "object": "fine_tuning.job",
  "id": "ftjob xxx",
  "model": "gpt-3.5-turbo-0613",
  "created_at": 1700626516,
  "finished_at": null,
  "fine_tuned_model": null,
  "organization_id": "org-zxxx",
  "result_files": [],
  "status": "validating_files",
  "validation_file": null,
  "training_file": "file-xxx",
  "hyperparameters": {
    "batch_size": "auto",
    "learning_rate_multiplier": "auto"
  },
  "trained_tokens": null,
  "error": null
}
have not n_epochs variable ,cause deserialize exception.

### DIFF
--- a/async-openai/src/types/types.rs
+++ b/async-openai/src/types/types.rs
@@ -836,7 +836,8 @@ pub struct Hyperparameters {
     ///
     /// "auto" decides the optimal number of epochs based on the size of the dataset.
     /// If setting the number manually, we support any number between 1 and 50 epochs.
-    pub n_epochs: NEpochs,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub n_epochs: Option<NEpochs>,
 }
 
 #[derive(Debug, Serialize, Clone, Default, Builder, PartialEq)]


64bit:
Hey thanks for PR, I'm not sure I follow what's going on. 

My guess is that `#[serde(skip_serializing_if = "Option::is_none")]` is missing attribute on `CreateFineTuningJobRequest` struct's `hyperparameters` field.

64bit:
Ah also looks like `Hyperparameters` struct needs updated for the spec:

```
        hyperparameters:
          type: object
          description: The hyperparameters used for the fine-tuning job.
          properties:
            batch_size:
              description: |
                Number of examples in each batch. A larger batch size means that model parameters
                are updated less frequently, but with lower variance.
              oneOf:
                - type: string
                  enum: [auto]
                - type: integer
                  minimum: 1
                  maximum: 256
              default: auto
            learning_rate_multiplier:
              description: |
                Scaling factor for the learning rate. A smaller learning rate may be useful to avoid
                overfitting.
              oneOf:
                - type: string
                  enum: [auto]
                - type: number
                  minimum: 0
                  exclusiveMinimum: true
              default: auto
            n_epochs:
              description: |
                The number of epochs to train the model for. An epoch refers to one full cycle 
                through the training dataset.
              oneOf:
                - type: string
                  enum: [auto]
                - type: integer
                  minimum: 1
                  maximum: 50
              default: auto
```

64bit:
Hmm there's perhaps confusion between deprecated and new fine tuning APIs.

inszee:
> Hey thanks for PR, I'm not sure I follow what's going on.
> 
> My guess is that `#[serde(skip_serializing_if = "Option::is_none")]` is missing attribute on `CreateFineTuningJobRequest` struct's `hyperparameters` field.

 There's not  impact on the structure of CreateFineTuningJobRequest。`#[serde(skip_serializing_if = "Option::is_none")]` is used to declare that a field in the structure may not exist during deserialization

PR #158, opened 2023-11-26
Fix: Make ChatCompletionToolChoiceOption enum variants lowercase

dmweis:
Setting this to `auto` was causing my requests to fail because of malformed values in the `tool_choice` filed.

I think the OpenAI API is expecting these values to be lowercase but "Auto" will be capitalised.
https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice

I only tested this with the "Auto" value. But I assume this should also work for the other variants


64bit:
Thank you for the fix!

PR #162, opened 2023-11-27
feat: transcribe with non-default response format

XUJiahua:
If transcribe with non-default response format like srt, the response is not a JSON.


64bit:
Thank you for the PR - this has been addressed in v0.20.0, hence closing.

PR #174, opened 2024-01-06
Optional model in ModifyAssistantRequest

CakeCrusher:
@64bit 
All fields (including model) are optional in OpenAI API.


64bit (approved):
Thank you for the fix!

PR #122, opened 2023-10-16
add basic wasm support

ifsheldon:
This is based on #120 and #121. 

To summarize:
* Added two feature gates `enable_tokio` and `enable_backoff`
   * `backoff` is somehow difficult to make it support wasm, so I just decided to gate it under a feature
* A minor API breaking change. Stream related functions now return `OpenAIEventStream` but it should not matter as long as users only use it in `while let await` loops since it also implements `Stream`
* Updated `reqwest-eventsource`

This should close #102 unless file-related ops are wanted.

I tested the code by
`cargo build --target wasm32-unknown-unknown --no-default-features --features wasm` and an web app.

The code is 

```rust
// main.rs
use async_openai::types::{ChatCompletionRequestMessageArgs, CreateChatCompletionRequestArgs, Role};
use dioxus::prelude::*;
use futures::stream::StreamExt;
use anyhow::Result;

const API_BASE: &str = "...";
const API_KEY: &str = "...";
const API_VERSION: &str = "...";
const DEPLOYMENT_ID: &str = "...";

pub fn app(cx: Scope) -> Element {
    let ok_count = use_state(cx, || 0_usize);
    let err_count = use_state(cx, || 0_usize);
    let response_string: &UseRef<String> = use_ref(cx, String::new);
    let fetch_completion_chunks: &Coroutine<()> = use_coroutine(cx, |rx| {
        let ok_count = ok_count.to_owned();
        let err_count = err_count.to_owned();
        let response_string = response_string.to_owned();
        async move {
            let config = async_openai::config::AzureConfig::new()
                .with_api_base(API_BASE)
                .with_api_key(API_KEY)
                .with_api_version(API_VERSION)
                .with_deployment_id(DEPLOYMENT_ID);
            let client = async_openai::Client::with_config(config);
            let request = CreateChatCompletionRequestArgs::default()
                .max_tokens(512u16)
                .model("gpt-3.5-turbo-0613")
                .messages([ChatCompletionRequestMessageArgs::default()
                    .role(Role::User)
                    .content("Hello!")
                    .build().unwrap()])
                .build().unwrap();
            let mut stream = client.chat().create_stream(request).await.unwrap();
            while let Some(chunk) = stream.next().await {
                match chunk {
                    Ok(response) => {
                        ok_count.modify(|x| *x + 1);
                        response_string.with_mut(|old| {
                            old.push('\n');
                            old.push_str(format!("{:?}", response).as_str());
                        })
                    }
                    Err(e) => {
[...]

PR #178, opened 2024-01-10
Add cloudflare wasm worker example

ifsheldon:
This adds a working cloudflare wasm worker example. 

A remaining issue is that this example must be moved out of the parent `async-openai` directory to get compiled, because `cargo` resolves dependencies of examples altogether, but `wasm-bindgen` causes conflicts between this example and openai-web-app example. This one uses older `wasm-bindgen` while the other requires newer one.

If the example is enough, you can also consider merging the code in experiments branch into main.

### DIFF
--- a/examples/cloudflare-wasm-worker/.gitignore
+++ b/examples/cloudflare-wasm-worker/.gitignore
@@ -1,0 +1,5 @@
+.idea
+target
+build
+node_modules
+.wrangler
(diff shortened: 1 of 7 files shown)


64bit:
Thank you for the PR. 

I want to remind you of expectations from my [comment](https://github.com/64bit/async-openai/pull/122#issuecomment-1802427031) . 

I'm sorry but I don't seem to have bandwidth to maintain this big of a change in `main` anytime in the near future. 

However, I'm happy to accept WASM related changes in `experiments` as long as you want to contribute. 

Thank you for your understanding again!

64bit (approved):
Not fully reviewed - accepting PR

Issue #177, opened 2024-01-08
Support text-to-speech (TTS) streaming functionality 


bqm:
**Problem**

The current `speech(...)` and `transcribe(...)` functions as part of the `Audio` implementation do not support a streaming mode.

This is particularly useful for any real-time application to simulate interactivity.

**Proposal**

Implement a `speech_stream` and `transcribe_stream`  mimicking the `create_stream` functionality.

Is someone already working on this? If not, I can give it a go.

64bit:
OpenAI doesn't seem to have native streaming APIs for Audio https://platform.openai.com/docs/api-reference/audio

Can you elaborate more on what you mean by streaming functionality for TTS & STT?

bqm:
From what I can tell, this might be a documentation oversight? https://help.openai.com/en/articles/8555505-tts-api mentions:

> Is it possible to stream audio?
Yes! By setting stream=True, you can chunk the returned audio file.

And people reports that it is indeed working for them: https://github.com/openai/openai-python/issues/864. It also matches the iOS app conversational behavior if you tried that.

There is also a merged PR for the python library: https://github.com/openai/openai-python/pull/866/commits/295581b674504433529f0bcaa7430f851ef990ff - although I did not try it myself.

I actually have not tried it myself yet - other than through postman which gives me a "Transfer-Encoding: chunked" header in the response so that might work out of the box without any specific "stream" key set to true. The rust library would need to expose that.

64bit:
Thank you for sharing additional information. The help article suggest that they have officially made it public and so I assume its safe to consider it as not a internal feature flag in their API.

Of course I'm not tackling it - as I learned it from your comment.

Given that, you're welcome to send a PR!  `stream` suffix in method name does sound reasonable as its consistent with other streaming APIs. 

In addition, having a working example for this would be very helpful for me to test and other folks to use.

64bit:
Upstream spec was updated after I had released v0.18.0. So perhaps it may have this or may not, but worth a look.

bqm:
Ok great, I should be able to push a PR tomorrow I think (for the speech_stream endpoint initially and can do a second PR for the transcribe endpoint after that?)

64bit:
Sounds like a good plan to me, thank you for offering to contribute!

bqm:
Just added a PR - I had a quick look for the STT use case - I don't think that streaming is actually supported for the transcription endpoint, looking at the OpenAI documentation and openai-python code.

[...]

PR #179, opened 2024-01-10
Add support for streaming TTS response from the /audio/speech endpoint

bqm:
**Problem**

OpenAI supports streaming for its /audio/speech endpoint but this functionality is not available in `async-openai`.

**Solution**

* Add support for response streaming following the existing coding patterns from the codebase.

A couple of decisions were taken:
- Add a `post_raw_stream` functionality in the client as using the existing SSE pattern would not work (the EventSource library decodes its data into a String and there is no byte mode from what I could tell).
- Do not introduce new specific types for the input and output structures, reusing the equivalent non stream type versions. This follows the existing convention from what I could tell.
- Introduce a simple example streaming to disk the audio output. Another example could have been to stream to the audio output but this would have needed the introduction of a playback library as well as a streaming audio decoder, which feels out of scope of this library. 

Related to #177.

### DIFF
--- a/async-openai/src/audio.rs
+++ b/async-openai/src/audio.rs
@@ -59,6 +59,7 @@ impl<'c, C: Config> Audio<'c, C> {
         let stream = Box::pin(self.client.post_raw_stream("/audio/speech", request)
             .await?
             .map(|stream_item| stream_item.map(|b| CreateSpeechResponse { bytes: b })));
+
         Ok(stream)
     }
 }
--- a/async-openai/src/audio.rs
+++ b/async-openai/src/audio.rs
@@ -1,9 +1,12 @@
+use futures::{stream::StreamExt, Stream};
+use std::pin::Pin;
 use crate::{
     config::Config,
     error::OpenAIError,
     types::{
         CreateSpeechRequest, CreateSpeechResponse, CreateTranscriptionRequest,
         CreateTranscriptionResponse, CreateTranslationRequest, CreateTranslationResponse,
+        SpeechResponseStream
     },
     Client,
 };
@@ -45,5 +48,17 @@ impl<'c, C: Config> Audio<'c, C> {
         let bytes = self.client.post_raw("/audio/speech", request).await?;
 
         Ok(CreateSpeechResponse { bytes })
+    }
+
+    /// Generates audio from the input text.
+    pub async fn speech_stream(
+        &self,
+        request: CreateSpeechRequest,
+    ) -> Result<SpeechResponseStream, OpenAIError> {
+
+        let stream = Box::pin(self.client.post_raw_stream("/audio/speech", request)
+            .await?
+            .map(|stream_item| stream_item.map(|b| CreateSpeechResponse { bytes: b })));
+        Ok(stream)
     }
 }
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -339,6 +339,29 @@ impl<C: Config> Client<C> {
         stream(event_source).await
     }
 
+    /// Make HTTP POST request to receive raw bytes for a stream based endpoint
+    pub(crate) async fn post_raw_stream<I>(
+        &self,
+        path: &str,
+        request: I,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Bytes, OpenAIError>> + Send>>, OpenAIError>
+    where
+        I: Serialize,
+    {
+        let response = self
+            .http_client
+            .post(self.config.url(path))
+            .query(&self.config.query())
+            .headers(self.config.headers())
+            .json(&request)
+            .send()
+            .await
+            .map_err(OpenAIError::Reqwest)?;
+
+        let stream = response.bytes_stream().map(|item| item.map_err(OpenAIError::Reqwest));
+        Ok(Box::pin(stream))
+    }
+
     /// Make HTTP GET request to receive SSE
     pub(crate) async fn get_stream<Q, O>(
         &self,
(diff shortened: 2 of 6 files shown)


async-openai/src/audio.rs
```diff
@@ -46,4 +49,17 @@ impl<'c, C: Config> Audio<'c, C> {
...
         Ok(CreateSpeechResponse { bytes })
     }
+
+    /// Generates audio from the input text.
+    pub async fn speech_stream(
+        &self,
+        request: CreateSpeechRequest,
+    ) -> Result<SpeechResponseStream, OpenAIError> {
+
+        let stream = Box::pin(self.client.post_raw_stream("/audio/speech", request)
```
64bit, line 59: Where is the `stream=True` provided? It seems that is equivalent to non-streaming in absence of the `stream=True`

examples/audio-speech-stream/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+        .build()?;
+
+    let mut stream = client.audio().speech_stream(request).await?;
+
+    let mut file = File::create("./speech_stream.mp3")?; 
+    
+    stream.for_each(|item| {
+        match item {
+            Ok(response) => {
+                let res_size = file.write(&response.bytes);
```
64bit, line 30: This example is not working as intended. If I dont provide API key I would expect error to log, but instead it writes the error json to mp3 file.

64bit: The example does work when provided the API key, but i"m still uncertain how without `stream=True` ?

async-openai/src/client.rs
```diff
@@ -339,6 +339,30 @@ impl<C: Config> Client<C> {
...
+            .http_client
+            .post(self.config.url(path))
+            .query(&self.config.query())
+            .headers(self.config.headers())
+            .json(&request)
+            .send()
+            .await
+            .map_err(OpenAIError::Reqwest)?;
+
+        let stream = response.bytes_stream().map(|item| item.map_err(OpenAIError::Reqwest));
```
64bit, line 361: My guess is that its working without `stream=True` because this converts the actual non-streaming response into byte stream?

bqm: See previous comment - from my tests, `stream=True` does not do anything... The response is streamed anyway from OpenAI side... If one has some doubts, one can confirm it by requesting a large text (> 2-3 paragraphs) and see the bytes being written to disk ~800ms  after the request.

Let me know if you cannot reproduce this

examples/audio-speech-stream/src/main.rs
bqm: Ah - this is my bad. I had assumed that a different status code from 200 would go in the Result::Err branch... but now I see that the conversion logic is part of the `execute_raw` function. I will move that code around so I can reuse it and ensure that I do some error tests as well.

async-openai/src/audio.rs
bqm: The `stream` parameter does not do anything from my tests - the response is streamed + chunked regardless. You can confirm it using curl / postman or something else. I also have a short example that playbacks the audio on the fly instead of writing it to disk - and that typically starts straightaway, I can add it to the example folder if that is helpful - I felt like it was slightly out of scope of the library but I don't mind adding it if that's helpful.

bqm:
@64bit I have updated my code to reflect your comment about errors. I have also used that opportunity to use a retry setup similar to the existing. The code can be further abstracted if that is desirable. I tested both error (no API key) and success scenario on my example and both are working well.

I have reused the map_err code from `execute_raw`, I can duplicate it or abstract the code even further - not sure what the usual setup is for this repository.

64bit:
Thanks for update. 

Given that previous change was not well tested and required major refactor, that calls for testing all existing examples. Are they all still working? 

Also based on new information the [assumption](https://github.com/64bit/async-openai/issues/177#issuecomment-1882091666) no longer holds, official document do not match what you have observed. 

Perhaps its best to wait until it makes it into OpenAPI spec.

bqm:
Thank you for reviewing @64bit.

> Perhaps its best to wait until it officially makes it into OpenAPI spec.
Ok let me know what stance you would like to have. I don't think the OpenAI spec will change soon but I am happy to report the discrepancy to them wherever is appropriate and wait if that is what you prefer.

I have a couple of potential exclusive suggestions:
1) Do not merge this change if you think it is best to wait until the OpenAI spec is updated.
2) Change this PR to minimize code changes, so essentially do not reuse the code, just duplicate the useful portions for now. This can give you piece of mind that if something is broken, it would just be the new code, not anything else.
3) Keep the PR as is and Do a lot of manual tests, both error path and happy path.

Let me know what is your preference - I am ok either way, happy to go with 1), 2) or 3). I do think streaming TTS is useful - people looking to use Rust are usually looking for performance imo and so this would make sense to have that feature at some point.

bqm:
@64bit Do you have any preferences between 1), 2) or 3)?

64bit:
Receiving a non-working PR / lack of testing has dropped my confidence in it.  

I dont have a preference at the moment.

bqm:
Hello, I am sorry to hear that. I am not sure how I can contribute at this point - I followed your initial requests, responded to your feedback and I am proposing various ways to move forward. Maybe let's drop this PR for now then, I will close it.

async-openai/src/client.rs
```diff
@@ -263,41 +263,88 @@ impl<C: Config> Client<C> {
                 .map_err(backoff::Error::Permanent)?;
 
             let status = response.status();
```
64bit, line 265: I think this line 265 is not used anymore, can be removed?

64bit:
Following up after your comment in the linked issue:

Left one comment, please let me know outcome of (3) and this is good to go.

PR #181, opened 2024-01-16
Borrow instead of moving CreateChatCompletionRequest

amosjyng:
I wish to retain control over the `CreateChatCompletionRequest` for logging and caching purposes after the call is made. Since the `create` function doesn't actually need to own any of the data in the request, it can simply borrow it instead.

This PR would unfortunately break the current API, but I'd like to bring it to your attention as a consideration for the next breaking release.

### DIFF
--- a/async-openai/Cargo.toml
+++ b/async-openai/Cargo.toml
@@ -40,9 +40,9 @@ tokio-stream = "0.1.11"
 tokio-util = { version = "0.7.7", features = ["codec", "io-util"] }
 tracing = "0.1.37"
 derive_builder = "0.12.0"
-async-convert = "1.0.0"
 secrecy = { version = "0.8.0", features=["serde"] }
 bytes = "1.5.0"
+async-trait = "0.1.52"
 
 [dev-dependencies]
 tokio-test = "0.4.2"
--- a/async-openai/src/assistant_files.rs
+++ b/async-openai/src/assistant_files.rs
@@ -27,7 +27,7 @@ impl<'c, C: Config> AssistantFiles<'c, C> {
     /// Create an assistant file by attaching a [File](https://platform.openai.com/docs/api-reference/files) to an [assistant](https://platform.openai.com/docs/api-reference/assistants).
     pub async fn create(
         &self,
-        request: CreateAssistantFileRequest,
+        request: &CreateAssistantFileRequest,
     ) -> Result<AssistantFileObject, OpenAIError> {
         self.client
             .post(&format!("/assistants/{}/files", self.assistant_id), request)
--- a/async-openai/src/assistants.rs
+++ b/async-openai/src/assistants.rs
@@ -30,7 +30,7 @@ impl<'c, C: Config> Assistants<'c, C> {
     /// Create an assistant with a model and instructions.
     pub async fn create(
         &self,
-        request: CreateAssistantRequest,
+        request: &CreateAssistantRequest,
     ) -> Result<AssistantObject, OpenAIError> {
         self.client.post("/assistants", request).await
     }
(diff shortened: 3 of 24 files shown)


64bit:
Thank you for the PR! 

While its a small change, it creates an inconsistent developer experience with remaining APIs that library exposes. 

Here are few options that I'd suggest:
1. Make all other APIs to accept a  reference (you're also welcome to create an issue to track it)
2. Clone `CreateChatCompletionRequest` to retain control (`Clone` trait is already derived on all types)

You're welcome to pick any choice from above.

amosjyng:
Sorry for the delay, I was trying to meet a deadline.

This looks to be a little harder than expected. If I do it for, say, this function in `Files`:

```rust
    pub async fn create(&self, request: &CreateFileRequest) -> Result<OpenAIFile, OpenAIError> {
        self.client.post_form("/files", request).await
    }
```

I get the error

```
error[E0277]: the trait bound `Form: async_convert::TryFrom<&types::file::CreateFileRequest>` is not satisfied
   --> async-openai\src\file.rs:26:41
    |
26  |         self.client.post_form("/files", request).await
    |                     ---------           ^^^^^^^ the trait `async_convert::TryFrom<&types::file::CreateFileRequest>` is not implemented for `Form`
    |                     |
    |                     required by a bound introduced by this call 
    |
```

If I then try to change that `TryFrom` to take in a reference instead:

```rust
#[async_convert::async_trait]
impl<'a> async_convert::TryFrom<&'a CreateFileRequest> for reqwest::multipart::Form {
    type Error = OpenAIError;

    async fn try_from(request: &'a CreateFileRequest) -> Result<Self, Self::Error> {
        let file_part = create_file_part(request.file.source).await?;
        let form = reqwest::multipart::Form::new()
            .part("file", file_part)
            .text("purpose", request.purpose);
        Ok(form)
    }
}
```

then this error crops up:

```
error[E0276]: impl has stricter requirements than trait
   --> async-openai\src\types\impls.rs:753:6
    |
753 | impl<'a> async_convert::TryFrom<&'a CreateFileRequest> for r... 
    |      ^^ impl has extra requirement `'a: 'async_trait`
```

This appears to be a [known issue](https://stackoverflow.com/a/69561872), and the [suggested solution](https://github.com/dtolnay/async-trait/issues/8#issuecomment-514812245) is hard to apply here because we can't edit the signature of the [`async_convert::TryFrom`](https://docs.rs/async-convert/1.0.0/src/async_convert/lib.rs.html#69-75) trait to support the `'async_trait` lifetime.

I've uploaded my edits thus far, but I don't know how feasible it is to do this for the async traits. What do you think about doing this just for the requests that don't require a `TryFrom`?

I'm ok with simply closing this PR too and maintaining my own fork. I just wanted to run the idea by you, that's all.

amosjyng:
> let say we do this for requests that don't require a TryFrom yet (required mostly for file uploads) and extrapolating that to future chat completion request which takes file uploads too - it would have similar problem and wont be future proof. Right?

Good point!

> Alternatively, if editing signature of async_convert::TryFrom is the only blocker and you'd like to explore further - [async-convert](https://github.com/yoshuawuyts/async-convert/blob/main/src/lib.rs) is very small library with MIT license, may be copy that (with honoring the license requirements) file into here and modify as needed? Or make upstream contributions to it and use here?

Sounds good. I've gotten the code compiling and tests passing. Some `Clone`'s are still needed unless we change the structs to take in a `Cow` string for reqwest. I had to change the test in `forks/async-openai/async-openai/tests/boxed_future.rs` because the Babbage model appears completely incapable of following any patterns anymore. Let me know what you think!

64bit:
With [Bring your own types](https://github.com/64bit/async-openai?tab=readme-ov-file#bring-your-own-types) feature, you should be able to pass references too as long as they meet the bounds requirements.

PR #185, opened 2024-01-29
Added dimensions param to embedding request

vmg-dev:
Added the dimensions param to the embedding request so that shortened embeddings can optionally be requested.


64bit (approved):
Thank you for the PR and nice and simple test!

PR #186, opened 2024-01-31
Sync experiments branch with main 0.18.2

ifsheldon:
This syncs the changes from main to experiments branch. The code compiles fine on all targets.


64bit (approved):
Thank you for keeping it in sync!

Issue #187, opened 2024-02-06
OpenAI  transcriptions api language field is invalid !


Taoaozw:
Because no value is assigned when it is converted to form data

64bit:
This is released in v0.18.3. Thank you for contributions!

PR #188, opened 2024-02-06
fix: CreateTranscriptionRequest language field not convert

Taoaozw:
When I was using the api `/v1/audio/transcriptions` , I found that the language conversion function was not working properly, and I fixed it! 😊 

issue is there: #187


64bit (approved):
Thank you for the bug fix!

PR #190, opened 2024-02-08
add support for base64 embeddings

### LINKED ISSUE #189: Embedding deserialization error when using encoding_format `EncodingFormat::Base64`
OpenAI returns embeddings in base64 string representation (offers better compactness than a JSON array) when specifying `EncodingFormat::Base64`. However, this is not handled in deserialization logic of `async-openai`.

Imo, Base64 should also be the default since much less data is transferred (as is the case in the python client).

Snippet:

```rs
async fn embed(openai: &Client<OpenAIConfig>, text: String) -> Result<Embedding> {
    let input: String = text.into();
    let request = CreateEmbeddingRequestArgs::default()
        .model("text-embedding-3-small")
        .input(input)
        .encoding_format(EncodingFormat::Base64)
        .build()
        .context("OpenAI embedder: failed to build text embedding request")?;
    let mut response = openai
        .embeddings()
        .create(request)
        .await
        .context("OpenAI embedder: failed to get text embedding")?;
    if response.data.len() != 1 {
        anyhow::bail!("Expected 1 embedding, got {}.", response.data.len());
    }
    Ok(response.data.remove(0).embedding.into())
}
```

adri1wald:
Resolves #189 

---

Considerations:

- Instead of using a generic could copy pasta the whole response struct for Base64 embeddings. This was less code, but leaks some implementation details out of the crate.
- Left the debug assert in - lmk if you want it removed
- Open to any other suggestions :)


64bit:
Thank you for the  PR! 

Looking at the spec doesn't distinguishes base64 response - it opens up implementation details - which you have done nicely!

I'll leave the comments about your considerations separately.

async-openai/src/types/embedding.rs
```diff
@@ -64,6 +64,17 @@ pub struct Embedding {
     pub embedding: Vec<f32>,
 }
 
+/// Represents an base64-encoded embedding vector returned by embedding endpoint.
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub(crate) struct Base64Embedding {
+    /// The index of the embedding in the list of embeddings.
+    pub(crate) index: u32,
```
64bit, line 71: Any reason that this is `pub(crate)` and not `pub`? 

Every other type exposes all the fields. May be this can be pub too if someone wants raw base64?

async-openai/src/embedding.rs
```diff
@@ -23,14 +27,56 @@ impl<'c, C: Config> Embeddings<'c, C> {
...
+    }
+}
+
+impl TryFrom<Base64Embedding> for Embedding {
+    type Error = OpenAIError;
+
+    fn try_from(embedding: Base64Embedding) -> Result<Self, Self::Error> {
+        let bytes = general_purpose::STANDARD.decode(embedding.embedding)?;
+        let chunks = bytes.chunks_exact(4);
+        debug_assert!(chunks.remainder().len() == 0);
```
64bit, line 64: Well this is nice.

If you're expecting this to ever fail in release build - perhaps return Err? 
 Otherwise its all good.

async-openai/src/types/embedding.rs
adri1wald: My initial thought was to hide the base64 encoding stuff from the end user. To them `EncodingFormat::Base64` "just speeds up the data transfer".

adri1wald: As is, this struct will never actually be returned to the end user, since we take care of conversion internally, so it made sense to me to make it `pub(crate)`. I think there are two approaches:

1. Make this `pub` and return the raw base64-encoded embeddings. Difficulty with this is, you end up with a sort of "overload", where part of the request param impacts the response type. You would either need two functions or return an enum to match over.
2. Hide completely the base64 stuff by removing the generic. The user never has to think about any of this.

async-openai/src/embedding.rs
adri1wald: I think this would be analogous to checking whether OpenAI returns an embedding that is of the correct dimensionality, or whether the response contains the correct number of embeddings. It "just won't happen". So I think it is probably safe to remove.

async-openai/src/types/embedding.rs
64bit: Thank you for elaborating. 

(2) makes sense, however, if user is choosing base64 response - the library should provide them exact response - which if they want can convert easily by `try_into()` that you have.  Hypothetically, if this PR doesn't exist user would still get same result - except the data transfer optimization that you have mentioned. 

Hence, (1) is favorable to get best of both world and gives user a choice - returns base64 response and lets them convert via try_into.  

For (1) you're right we need two functions - we already do this for stream and non-streaming counter parts for chat completion API (and other streaming apis).

Something like  `Embeddings::create` and `Embeddings::create_base64` to make it explicit.

64bit: In addition, two functions also gives you leverage to make it a non-breaking change by having a new type name for base64 counterpart.

64bit: Hey just checking in - if there's any feedback/followup about the discussion?

adri1wald: Hey sorry I forgot about this. Will have a go tonight.

64bit: Oh no worries, thank you for following up, please take your time

adri1wald:
@64bit Sorry about the delay. I think this is ready for another review now! Let me know what you think :)

64bit (approved):
This is looking good! thank you for the updates and including tests!

PR #195, opened 2024-03-01
Add usage information to the run object

turingbuilder:
OpenAI provide usage information on the run object, this PR adds that information to the RunObject in async_openai.

OpenAI documentation:
https://platform.openai.com/docs/api-reference/runs/object#runs/object-usage


64bit (approved):
Thank you for the PR!

The names & doc comments should match the spec -  I'll take care of it - just heads up if you see changes before I publish this.

PR #197, opened 2024-03-02
Add Client::build for full customizability during instantiation

### LINKED ISSUE #193: Initialize `Client` with custom HTTP client, config, and backoff
Initializing a `Client` with a custom HTTP client always requires first creating a useless `reqwest::Client` in `Client::with_config`.

This adds a bit of useless overhead due to `reqwest`'s  TLS/DNS functionality, which can be non-trivial for certain low-latency applications.

Would you be open to adding an additional method along the lines of?

```rs
impl<C: Config> Client<C> {
    /// Create client with your own HTTP client, config, and backoff 
    pub fn build(http_client: reqwest::Client, config: C, backoff: backoff::ExponentialBackoff) -> Self {
        Self {
            http_client,
            config,
            backoff,
        }
    }

...
```

GabrielBianconi:
Fixes #193

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -35,6 +35,19 @@ impl Client<OpenAIConfig> {
 }
 
 impl<C: Config> Client<C> {
+    /// Create client with a custom HTTP client, OpenAI config, and backoff.
+    pub fn build(
+        http_client: reqwest::Client,
+        config: C,
+        backoff: backoff::ExponentialBackoff,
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you for the PR! 

I think the reason for this change is mostly in the issue conversation - a doc line it could help why we have this if someone is looking at docs.rs or just browsing code - just a thought on documentation.

PR #200, opened 2024-03-13
Making the sleep asynchronous compatible in an example

### LINKED ISSUE #194: Usage of std::thread::sleep inside async function in assistants example
In this [line](https://github.com/64bit/async-openai/blob/48d18e02dc8afa551cf68f7676bceda395f2f9b1/examples/assistants/src/main.rs#L145) in the assistants example, you are using std::thread::sleep - this would block the whole async runtime

sgopalan98:
Resolves #194 .

Using `std::thread::sleep` in async code is not advisable. `tokio::time::sleep` is preferred instead. This PR contains this small change. 

Tested the code change by running locally - works as expected!

### DIFF
--- a/examples/assistants/src/main.rs
+++ b/examples/assistants/src/main.rs
@@ -135,7 +135,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                 }
             }
             //wait for 1 second before checking the status again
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you for the fix and testing!

Issue #201, opened 2024-03-14
Missing fields in `CreateTranscriptionResponse`


emk:
Thank you for such a great library! I've been experimenting with this in an open source tool.

Normally, if I set `timestamp_granularities=["segment", "word"]` when creating a transcription, the API returns:

```json
{
    "language": "spanish",
    "duration": 1409.5999755859375,
    "text": "...",
    "words": [
        {
            "word": "...",
            "start": 2.4800000190734863,
            "end": 3.1600000858306885
        }
    ],
    "segments": [
        {
            "id": 0,
            "seek": 0,
            "start": 2.4800000190734863,
            "end": 5.860000133514404,
            "text": "...",
            "tokens": [
                50414,
                ...
            ],
            "temperature": 0.20000000298023224,
            "avg_logprob": -0.28230372071266174,
            "compression_ratio": 1.465000033378601,
            "no_speech_prob": 0.22726023197174072
        }
    ]
}
```

But the current release only supports `text`, not the other fields.

I might be able to put together a PR when I get a minute. Thank you again for a great library!

64bit:
Thank you for contributions! the changes for this is released in v0.20.0

emk:
Fantastic! Thank you.

Issue #163, opened 2023-12-02
Auto Generation from OpenAPI spec


64bit:
**Background** 

This project started around Nov 2022 because I did not find any production ready Rust libraries for OpenAI covering all APIs.

And my non-comprehensive search back then did not lead any robust tools for generating Rust code from spec, and so this crate was hand written. Which is not sustainable when OpenAI does big releases like DevDay.

This issue is to have discussions, and potential long term solutions for generating Rust from OpenAPI spec.  Something which helps not just `async-openai` but overall Rust eco-system for APIs with no production ready Rust crates  - like DigitalOcean or Linode OpenAPI spec etc.

louis030195:
this works: `openapi-generator generate -i openai-openapi.yaml -g rust -o ./output --skip-validate-spec`

openai seems to use https://www.stainlessapi.com/ but it does seem a commercial product

64bit:
Thanks for sharing the link.

Do you by any chance happen to have output of the `openapi-generator` ? I explored it back then but didn't work well

izyuumi:
@64bit I am interested in this enhancement. Here's the output from the `openapi-generator` https://github.com/izyuumi/openaiapi.

dynamite-bud:
@64bit @izyuumi .
I think the generator was helpful to generate structures and models. 

But I don't think they provide a very nice public facing API that can be used.

64bit:
Using agents these days to update: https://github.com/64bit/async-openai/blob/main/AGENTS.md

wilsonmar:
https://www.stainlessapi.com/ is gone. 
Is it now
https://www.stainless.com/
which say Anthropic is purchasing it.

PR #202, opened 2024-03-16
Update Audio APIs from updated spec

emk:
This PR adds support for `AudioResponseFormat::VerboseJson` and `TimestampGranularity`, including updated example code. These were defined as types before, but not fully implemented.

Implements 64bit/async-openai#201.

Please let me know if I can improve this PR! And thank you for a very helpful crate.

### DIFF
--- a/async-openai/src/audio.rs
+++ b/async-openai/src/audio.rs
@@ -1,9 +1,11 @@
+use bytes::Bytes;
+
 use crate::{
     config::Config,
     error::OpenAIError,
     types::{
         CreateSpeechRequest, CreateSpeechResponse, CreateTranscriptionRequest,
-        CreateTranscriptionResponse, CreateTranslationRequest, CreateTranslationResponse,
+        CreateTranscriptionResponseJson, CreateTranscriptionResponseVerboseJson, CreateTranslationRequest, CreateTranslationResponse,
     },
     Client,
 };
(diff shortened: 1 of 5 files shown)


64bit:
Hello, 

Thank you for the issue and the implementation! I'll leave review in a separate comment

64bit:
It seems they've updated spec for this. 

While logistically this implementation looks to fill the gap - however it doesn't adhere to the new spec - all doc comments, type names must match the spec.  

A quick look on spec suggest there are multiple new types (`CreateTranscriptionResponseJson`, `CreateTranscriptionResponseJsonVerbose`, etc.)  and this PR may require a bit more work to incorporate them all.

emk:
> While logistically this implementation looks to fill the gap - however it doesn't adhere to the new spec - all doc comments, type names must match the spec.

Ah, great! Do you have a tool to auto-generate this code from the spec, or are you synchronizing it by hand?

I was [working from these documents here](https://platform.openai.com/docs/api-reference/audio/createTranscription) originally, because they contained quite a bit of information that was missing elsewhere.

64bit:
Unfortunately no tool  to auto generate and its written by hand, I have full context about it here https://github.com/64bit/async-openai/issues/163

64bit:
> I was [working from these documents here](https://platform.openai.com/docs/api-reference/audio/createTranscription) originally, because they contained quite a bit of information that was missing elsewhere.

They also seem to generate website from the spec - and so I go with spec as source of truth.

64bit:
Thank you for the detailed choices. Those are all reasonable options. 

>  because the API chooses between them based on the function arguments

For this a pattern is already used in the crate:  to have separate functions to return separate types based on different value to a parameter to same request object. 

For example, please take a look at `Chat::create` and `Chat::create_stream`, and a more recent addition of `Embedding::create` and `Embedding::create_base64` . Same pattern can be applied here too.

emk:
OK, here's an improved version! I focused on two things:

1. Making the API as consistent as possible with the spec, and
2. After satisfying (1), trying to make as few breaking changes as possible.

Here are the specific changes I made:

- Modify comments and type names to match spec as precisely as possible.
- BREAKING CHANGE: Rename `CreateTranscriptionRespose` to `CreateTranscriptionResponseJson` (to match latest API spec)
- Add `CreateTranscriptionResponseVerboseJson` and `transcribe_verbose_json`
- Add `transcribe_raw` for SRT output, which is poorly documented in the OpenAPI spec.
- Add `post_form_raw`
- Update example code

I'm pretty sure this could be improved in various ways by making different tradeoffs, but I hope it's closer. Thank you for your feedback and for a very helpful library!

64bit:
> I've gotten caught up with emergency work stuff this weekend, and haven't had a chance to redo this PR. It's still very high on my list, and I hope to get to it ASAP!

No worries, unlike work, here, there's is no pressure or deadline or emergency.  Thank you for taking time to contribute!

64bit (approved):
This is good, thank you for matching with spec and adding support for all format! 

Your observation about raw/SRT etc. is worth documenting as doc comments for these new function(s) because then it becomes available for everyone on docs.rs, I might make edits for it before publishing.

emk:
Please do feel free to edit the doc comments! I was trying to stick extremely strictly to OpenAPI spec's comments, but it would very likely help users to add more information to the `transcribe_raw` function in particular.

Issue #134, opened 2023-11-07
Support gpt-4-vision-preview


sloganking:
https://platform.openai.com/docs/models/gpt-4-and-gpt-4-turbo

I imagine gpt-4 vision is going to have some API differences compared to normal language models that just take text instead of files. Supporting it would probably be desirable to many people however.

64bit:
Hey, just published DevDay released on v0.16.0. Please give it a try, if you find any issues PRs are welcome!

pecet:
So is vision supported yet, or only TTS?

ludberg:
How would we add something like this?
![image](https://github.com/64bit/async-openai/assets/3208326/58fc4ec2-6aa7-4473-b8c9-3e305e6b57c9)

I think I've found the corresponding code - do you have an example on how to use it?
![image](https://github.com/64bit/async-openai/assets/3208326/0b47c3b8-5597-4d0c-8866-d4d9e593ce38)

tiderjian:
[...]

PR #203, opened 2024-03-18
Add vision-chat example

Gabriel2409:
The example directory currently lacks an example that matches the quickstart from https://platform.openai.com/docs/guides/vision 

This PR adds the quickstart example. Notably it shows how to use `ChatCompletionRequestMessageContentPartTextArgs`, `ChatCompletionRequestMessageContentPartImageArgs`, `ImageUrlArgs` and `ImageUrlDetail`. 

I followed the logic from the standard chat example.
Please let me know if I can improve the PR and thank you very much for creating this crate.

### DIFF
--- a/examples/vision-chat/Cargo.toml
+++ b/examples/vision-chat/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "vision-chat"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+async-openai = { path = "../../async-openai" }
+serde_json = "1.0.114"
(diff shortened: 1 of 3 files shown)


64bit (approved):
Thank you for the vision example!

Looking at the nested `build()?`s suggest to me that library can be improved by implementing something like `From<S: Into<String>>` trait for `ChatCompletionRequestMessageContentPartText`, `ChatCompletionRequestMessageContentPartImage`, and  `ImageUrl`. 

If there's a default/single path from string family to the Rust struct for a type in this crate - From trait would make final code less verbose. 

Just a note for me for future library improvements, but you're welcome to explore it.

64bit:
Also closes https://github.com/64bit/async-openai/issues/134

PR #213, opened 2024-04-10
Sync wasm experiment branch to 0.20

ifsheldon:
This syncs the updates from main to experiments branch. 

Thanks to #154 , wasm support should be complete and mature enough to get merged into main branch.


64bit:
As always - thank you for keeping wasm branch latest!

PR #222, opened 2024-05-08
fix FileCitation deserialization

kitalia:
Made `pub quote: Option<String>`, optional as suggested, seems to work fine now, cheers


64bit:
Thank you for the fix ( i have taken liberty to update title)

64bit (approved):

kitalia:
Thank you for merging 👍

PR #226, opened 2024-05-21
assistant: don't send null values for tool resources

domenkozar:
Otherwise the server fails with 500.


64bit (approved):
Thank you for the bug fix and PR!

PR #230, opened 2024-06-06
fix: Update vector store file chunking strategy to use StaticChunkingStrategy

### LINKED ISSUE #229: Failing to deserialize `VectorStoreFileObject`
Error:
```
2024-06-06T04:57:33.665329Z ERROR async_openai::error: failed deserialization of: {
  "id": "file-cYrRWFomydf8Ng9gvqs4zWBD",
  "object": "vector_store.file",
  "usage_bytes": 0,
  "created_at": 1717649854,
  "vector_store_id": "vs_gN12Na14YvsPhjPPxS91WX3b",
  "status": "in_progress",
  "last_error": null,
  "chunking_strategy": {
    "type": "static",
    "static": {
      "max_chunk_size_tokens": 800,
      "chunk_overlap_tokens": 400
    }
  }
}
```

Failing code:
```rust
client
   .files()
   .create(CreateFileRequest {
       file: FileInput::from_vec_u8("meoww.txt".into(), memory.clone().into_bytes()),
       purpose: FilePurpose::Assistants,
   }))
   .expect("Failed to upload memory as file!");
```

Looks like the issue is [here](https://docs.rs/async-openai/latest/src/async_openai/types/assistant.rs.html#67):
```rust
/// Static Chunking Strategy
pub struct StaticChunkingStrategy {
    /// The maximum number of tokens in each chunk. The default value is `800`. The minimum value is `100` and the maximum value is `4096`.
    max_chunk_size_tokens: u16,
    /// The number of tokens that overlap between chunks. The default value is `400`.
    ///
    /// Note that the overlap must not exceed half of `max_chunk_size_tokens`.
    chunk_overlap_tokens: u16,
}
```

I'm new to Rust, but it seems like the `max_chunk_size_tokens` and `chunk_overlap_tokens` fields may need to be `pub`?

hiibolt:
This targets the changes mentioned in #229, because OpenAI requires a non-null name when creating a vector store. This also fixes attaching a file to a vector store, where it would fail to deserialize. Also adds a test for both errors to catch them during development in the future!

Fixes #229

### DIFF
--- a/async-openai/src/types/vector_store.rs
+++ b/async-openai/src/types/vector_store.rs
@@ -18,8 +18,7 @@ pub struct CreateVectorStoreRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub file_ids: Option<Vec<String>>,
     /// The name of the vector store.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
+    pub name: String,
 
     /// The expiration policy for a vector store.
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -192,7 +191,7 @@ pub enum VectorStoreFileErrorCode {
 pub enum VectorStoreFileObjectChunkingStrategy {
     /// This is returned when the chunking strategy is unknown. Typically, this is because the file was indexed before the `chunking_strategy` concept was introduced in the API.
     Other,
-    Static(StaticChunkingStrategy),
+    Static{ r#static: StaticChunkingStrategy },
 }
 
 #[derive(Debug, Serialize, Default, Clone, Builder, PartialEq)]
--- a/async-openai/src/types/vector_store.rs
+++ b/async-openai/src/types/vector_store.rs
@@ -18,7 +18,8 @@ pub struct CreateVectorStoreRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub file_ids: Option<Vec<String>>,
     /// The name of the vector store.
-    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
 
     /// The expiration policy for a vector store.
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -62,7 +63,7 @@ pub struct VectorStoreObject {
     /// The Unix timestamp (in seconds) for when the vector store was created.
     pub created_at: u32,
     /// The name of the vector store.
-    pub name: String,
+    pub name: Option<String>,
     /// The total number of bytes used by the files in the vector store.
     pub usage_bytes: u64,
     pub file_counts: VectorStoreFileCounts,
--- a/async-openai/src/vector_store_files.rs
+++ b/async-openai/src/vector_store_files.rs
@@ -107,7 +107,13 @@ mod tests {
                 metadata: None
             })
             .await?;
+    let vector_store_file = client
+        .vector_stores()
+        .files(&vector_store_handle.id)
+        .retrieve(&file_handle.id)
+        .await?;
 
+    assert_eq!(vector_store_file.id, file_handle.id);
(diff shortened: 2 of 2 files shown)


async-openai/src/types/vector_store.rs
```diff
@@ -18,8 +18,7 @@ pub struct CreateVectorStoreRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub file_ids: Option<Vec<String>>,
     /// The name of the vector store.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
```
64bit, line 22: Officially this is still optional, what's the reason for making it non-optional?

https://platform.openai.com/docs/api-reference/vector-stores/create#vector-stores-create-name

async-openai/src/vector_store_files.rs
```diff
@@ -75,3 +75,58 @@ impl<'c, C: Config> VectorStoreFiles<'c, C> {
...
+                    String::from(":3").into_bytes()
+                ),
+                purpose: FilePurpose::Assistants
+            }).await?;
+
+        // Create a vector store
+        let vector_store_handle = client
+            .vector_stores()
+            .create( CreateVectorStoreRequest {
+                file_ids: None,
```
64bit, line 103: You can actually attach `file_handle` in the request without a name too and it will work. That way name is still optional 

https://github.com/64bit/async-openai/issues/229#issuecomment-2155615304

async-openai/src/vector_store_files.rs
```diff
@@ -75,3 +75,58 @@ impl<'c, C: Config> VectorStoreFiles<'c, C> {
...
+                name: String::from("meowww"),
+                expires_after: None,
+                chunking_strategy: None,
+                metadata: None
+            })
+            .await?;
+
+        // Attach the file to the vector store
+        client
```
64bit, line 112: So once you provide `file_ids` during creation, this becomes redundant.

async-openai/src/types/vector_store.rs
hiibolt: Reverted in [next commit](https://github.com/64bit/async-openai/pull/230/commits/d368ce06fd470dec498bcb97863436b36b33416d)

async-openai/src/vector_store_files.rs
hiibolt: Modified in [following commit](https://github.com/64bit/async-openai/pull/230/commits/29a790a10774f017e9c46cef32817f66e2191fe0)

async-openai/src/vector_store_files.rs
hiibolt: Modified in [following commit](https://github.com/64bit/async-openai/pull/230/commits/29a790a10774f017e9c46cef32817f66e2191fe0)

async-openai/src/vector_store_files.rs
```diff
@@ -75,3 +75,49 @@ impl<'c, C: Config> VectorStoreFiles<'c, C> {
...
+            .vector_stores()
+            .create( CreateVectorStoreRequest {
+                file_ids: Some(vec![file_handle.id.clone()]),
+                name: None,
+                expires_after: None,
+                chunking_strategy: None,
+                metadata: None
+            })
+            .await?;
+
```
64bit, line 110:
```suggestion
    let vector_store_file = client
        .vector_stores()
        .files(&vector_store_handle.id)
        .retrieve(&file_handle.id)
        .await?;

    assert_eq!(vector_store_file.id, file_handle.id);
```

To test your actual first reported error for `VectorStoreFileObject` :)

hiibolt: Sounds good, thank you!

64bit (approved):
Thank you for the PR and bug fixes!

Issue #232, opened 2024-06-11
CreateChatCompletionRequest.max_tokens should be u32


MakotoE:
[`CreateChatCompletionRequest.max_tokens`](https://docs.rs/async-openai/latest/async_openai/types/struct.CreateChatCompletionRequest.html#structfield.max_tokens) is a `Option<u16>` as of 0.23.1.

The newer models such as [gpt-4o](https://platform.openai.com/docs/models/gpt-4o) has a context window of 128,000 tokens. This context window limit is the sum of input and output tokens.

I believe the `max_tokens` field should be `Option<u32>` to allow numbers as high as 128,000.

64bit:
Thank you for reporting the bug, a PR is most welcome!

PR #233, opened 2024-06-11
Change max_tokens type to Option<u32>

MakotoE:
Closes https://github.com/64bit/async-openai/issues/232


64bit (approved):
Nice! 

Thank you for a thorough fix!

Issue #234, opened 2024-06-14
Error when using assistant


madchicken:
I am getting an error when trying to start a thread for an assistant:

```
Error: ApiError(ApiError { message: "Invalid type for 'parallel_tool_calls': expected a boolean, but got null instead.", type: Some("invalid_request_error"), param: Some("parallel_tool_calls"), code: Some("invalid_type") })
```

I think OpenAI changed something VERY recently, since yesterday this was not happening.
The example under `assistants` is failing with the same error

64bit:
Oh no, thats no good. 

Thank you for reporting the issue.

From the error message you have provided, it seems that skipping serialization if parallel_tool_calls is None would prevent sending null to their API and fix it?

madchicken:
Please, disregard my previous message, the error is way below, when running the thread request:
```
        //create a run for the thread
        let run_request = CreateRunRequestArgs::default()
            .assistant_id(assistant_id)
            .build()?;
```
and I was able to fix it by doing what you said:
```
        //create a run for the thread
        let run_request = CreateRunRequestArgs::default()
            .assistant_id(assistant_id)
            .parallel_tool_calls(false)
            .build()?;
```

64bit:
Thank you for the issue and PR, the fix is released in v0.23.3

PR #235, opened 2024-06-16
fix: skip serialization to null if field is none in `CreateRunRequest`

madchicken:
Fix for issue #234

### DIFF
--- a/async-openai/src/types/run.rs
+++ b/async-openai/src/types/run.rs
@@ -192,45 +192,57 @@ pub struct CreateRunRequest {
     pub model: Option<String>,
 
     /// Overrides the [instructions](https://platform.openai.com/docs/api-reference/assistants/createAssistant) of the assistant. This is useful for modifying the behavior on a per-run basis.
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


async-openai/src/types/run.rs
```diff
@@ -228,7 +228,8 @@ pub struct CreateRunRequest {
     pub tool_choice: Option<AssistantsApiToolChoiceOption>,
 
     /// Whether to enable [parallel function calling](https://platform.openai.com/docs/guides/function-calling/parallel-function-calling) during tool use.
-    pub parallel_tool_calls: Option<bool>,
+    #[serde(default)]
+    pub parallel_tool_calls: bool,
```
64bit, line 232: This may work, however it has a semantics of a "required" param, which is not the source of truth - OpenAI API reference and spec says - its not a required parameter. 

I think core of the issue is not having `#[serde(skip_serializing_if = "Option::is_none")]` on all the `Option`-al  fields of `CreateRunRequest`  - if this attribute fixes the issues its more inline with semantics and consistency.

64bit (approved):
Thank you for the fix and PR!

PR #240, opened 2024-07-08
Make error message for `ApiError` more user-friendly

### LINKED ISSUE #238: Format `OpenAIError::APIError` with client-friendly message
In `error.rs`, we have:

```rs
...
    #[error("{:?}: {}", .0.r#type, .0.message)]
    ApiError(ApiError),
...
pub struct ApiError {
    pub message: String,
    pub r#type: Option<String>,
    pub param: Option<String>,
    pub code: Option<String>,
}
...
```

The `{:?}` leads to error messages that aren't client-friendly, such as:

```
Some(\"invalid_request_error\"): This model's maximum context length is 16385 tokens. However, your messages resulted in 25015 tokens (24961 in the messages, 54 in the functions). Please reduce the length of the messages or functions.
```

Could we replace it with something more client-friendly like:

```rs
...
    #[error("{}: {}", .0.r#type. unwrap_or("unknown_error_type".to_string()), .0.message)]
    ApiError(ApiError),
...
```

Thanks!

GabrielBianconi:
Fixes #238 

I tested with one variation ("too many tokens") and it works fine. Tough to test exhaustively of course.

### DIFF
--- a/async-openai/src/error.rs
+++ b/async-openai/src/error.rs
@@ -37,7 +37,7 @@ pub struct ApiError {
 }
 
 impl std::fmt::Display for ApiError {
-    /// If all fields are available, ApiError is formatted as:
+    /// If all fields are available, `ApiError` is formatted as:
     /// `{type}: {message} (param: {param}) (code: {code})`
     /// Otherwise, missing fields will be ignored.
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you for the PR, this is better than previous error message!

PR #241, opened 2024-07-08
derive Deserialize anywhere Serialize is derived

kczimm:
I'd like to deserialize into several structs that do not derive `Deserialize`. This PR derives `Deserialize` anywhere a `struct` derives `Serialize`.

I also fixed some clippy lints.


64bit (approved):
Nice, thank you for the PR!

PR #246, opened 2024-07-23
Add log compatibility

Crazytieguy:
See https://docs.rs/tracing/latest/tracing/#emitting-log-records


64bit:
Thank you for the PR!

Shouldn't this be a feature flag in https://github.com/64bit/async-openai/blob/main/async-openai/Cargo.toml#L15-L24 to opt in explicitly? May be `tracing-log = [ "tracing/log" ]` 

Since the library is not using the feature flag directly, it doesn't need to be enabled by default. Open to hear why you'd like to enable it by default?

Crazytieguy:
Oh, that sounds reasonable actually. Although in that case maybe tracing itself should also be enabled by a feature? I tried looking for examples in libraries I expected to have something like this, but apparently they don't have a way to enable tracing-log either. I wonder if I'm missing something.

Anyway it's your decision of course

PR #248, opened 2024-07-24
add methods to run chat and completion with extra args

kczimm:
This PR adds methods to the Chat and Completions API to allow serializing extra arguments into the payload for services that provide support for additional arguments (e.g. [vLLM](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#extra-parameters)).

### DIFF
--- a/async-openai/src/chat.rs
+++ b/async-openai/src/chat.rs
@@ -1,9 +1,12 @@
+use serde::Serialize;
+
 use crate::{
     config::Config,
     error::OpenAIError,
     types::{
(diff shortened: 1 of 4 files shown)


64bit:
Thank you, a PR is always appreciated, however this is out of the scope for this library, for more details please see [Contribution](https://github.com/64bit/async-openai?tab=readme-ov-file#contributing) guidelines

kczimm:
Ah, my apologies. I didn't read that. Thanks!

Issue #251, opened 2024-07-30
CreateAssistantToolResources fields must be optional?


vrurg:
I was working on code that creates an assistant with attached pre-existing vector store. Unfortunately, it doesn't seem to be possible. Since both fields are currently non-`Option`, the only possible way to create an instance is something like:

```rust
        let tool_resources = Some(CreateAssistantToolResources {
            file_search: Some(CreateAssistantToolFileSearchResources {
                vector_store_ids,
                vector_stores: vec![],
            }),
            code_interpreter: None,
        });
```

The point is that one has to specify both `vector_store_ids` and `vector_stores`. Unfortunately, the API doesn't accept this kind of request:

```
Can't create assistant '{aname}': ApiError(ApiError { message: "file_search tool configuration is invalid. Only one of vector_store_ids or vector_stores should be provided.", type: Some("invalid_request_error"), param: Some("tool_resources.file_search"), code: None })
```

From the error message I conclude that the right way is to make both fields optional.

Now I wonder how do I attach the vector store to a new assistant...

PR #252, opened 2024-07-30
Fix incompatibility with API requirements

vrurg:
For `CreateAssistantToolFileSearchResources` struct the API expects only one of the two fields to be used. Even an empty array is not accepted if the other field contains a value. Thus making both `vector_store_ids` and `vector_stores` optional is the way to go.

Fixes 64bit/async-openai#251


vrurg:
Luckily, this fix is very simple. :)

human-d3v (approved):
When I changed this locally, it fixed the API requirement issue.

64bit (approved):
Thank you for the PR!

PR #254, opened 2024-08-10
Add strict to function calls

linusbierhoff:
Added the `strict : <bool>` option for function calls to enable structured output. See [Structured Output](https://openai.com/index/introducing-structured-outputs-in-the-api/)


linusbierhoff:
No it’s basically the same

Issue #216, opened 2024-04-21
The type of messages in deserialized CreateChatCompletionRequest  are all SystemMessage


sontallive:
I want to deserialize request json to `CreateChatCompletionRequest` but i found the messages are all System. 

code
```rust
use async_openai::types::{
    ChatCompletionRequestSystemMessageArgs, ChatCompletionRequestUserMessageArgs,
    CreateChatCompletionRequest, CreateChatCompletionRequestArgs,
};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let request: CreateChatCompletionRequest = CreateChatCompletionRequestArgs::default()
        .messages([
            ChatCompletionRequestSystemMessageArgs::default()
                .content("your are a calculator")
                .build()?
                .into(),
            ChatCompletionRequestUserMessageArgs::default()
                .content("what is the result of 1+1")
                .build()?
                .into(),
        ])
        .build()?;
    // serialize the request
    let serialized = serde_json::to_string(&request)?;
    println!("{}", serialized);
    // deserialize the request
    let deserialized: CreateChatCompletionRequest = serde_json::from_str(&serialized)?;
    println!("{:?}", deserialized);
    Ok(())
}
```

result
```bash
{"messages":[{"content":"your are a calculator","role":"system"},{"content":"what is the result of 1+1","role":"user"}],"model":""}

CreateChatCompletionRequest { messages: [System(ChatCompletionRequestSystemMessage { content: "your are a calculator", role: System, name: None }), System(ChatCompletionRequestSystemMessage { content: "what is the result of 1+1", role: User, name: None })], model: "", frequency_penalty: None, logit_bias: None, logprobs: None, top_logprobs: None, max_tokens: None, n: None, presence_penalty: None, response_format: None, seed: None, stop: None, stream: None, temperature: None, top_p: None, tools: None, tool_choice: None, user: None, function_call: None, functions: None }
```

djmango:
I also have this issue. Using actix_web

djmango:
Was banging my head on this for a bit, but just pushed a fix on my branch.

thanks coco.codes from the NAMTAO discord!

to solve the parent issue, of them always being System, we implement the macro `#[serde(tag = "role", rename_all = "lowercase")]` in `ChatCompletionRequestMessage`

[...]

PR #255, opened 2024-08-12
Support structured outputs

### LINKED ISSUE #253: Add support for the structured output response format
OpenAI recently introduced the [Structured Outputs](https://openai.com/index/introducing-structured-outputs-in-the-api/) feature which ensures that responses will adhere to a predefined JSON schema.
> Last year at DevDay, we introduced JSON mode—a useful building block for developers looking to build reliable applications with our models. While JSON mode improves model reliability for generating valid JSON outputs, it does not guarantee that the model’s response will conform to a particular schema. Today we’re introducing Structured Outputs in the API, a new feature designed to ensure model-generated outputs will exactly match JSON Schemas provided by developers.

The documentation for using structured outputs is [here](https://platform.openai.com/docs/guides/structured-outputs).

domenkozar:
Fixes #253 

~I'm not entirely sure why the whole file is re-added.~


nytopop:
> I'm not entirely sure why the whole file is re-added.

The line endings have changed from `LF` to `CRLF`. `git diff -R` will show the change.

It looks like there's some inconsistency on that front, might be worth a separate issue/PR to cleanup:

```
eric@io ~/src/code/rust/async-openai/async-openai/src/types (response_format_json_schema)> file *
assistant.rs:        ASCII text, with very long lines (502)
assistant_file.rs:   ASCII text
assistant_impls.rs:  ASCII text
assistant_stream.rs: ASCII text
audio.rs:            ASCII text, with very long lines (353), with CRLF line terminators
batch.rs:            ASCII text
chat.rs:             ASCII text, with very long lines (506)
common.rs:           ASCII text, with CRLF line terminators
completion.rs:       Unicode text, UTF-8 text, with very long lines (569), with CRLF line terminators
embedding.rs:        ASCII text, with very long lines (460), with CRLF line terminators
file.rs:             ASCII text, with very long lines (388)
fine_tuning.rs:      ASCII text, with CRLF line terminators
image.rs:            ASCII text, with CRLF line terminators
impls.rs:            ASCII text
message.rs:          ASCII text
message_file.rs:     ASCII text
mod.rs:              ASCII text
model.rs:            ASCII text, with CRLF line terminators
moderation.rs:       ASCII text, with very long lines (344), with CRLF line terminators
run.rs:              ASCII text, with very long lines (358)
step.rs:             ASCII text
thread.rs:           ASCII text, with very long lines (358)
vector_store.rs:     ASCII text
```

async-openai/src/types/assistant.rs
```diff
@@ -129,8 +129,11 @@ pub enum AssistantsApiResponseFormatOption {
 /// An object describing the expected output of the model. If `json_object` only `function` type `tools` are allowed to be passed to the Run. If `text` the model can return text or any value needed.
 #[derive(Clone, Serialize, Debug, Deserialize, PartialEq, Default)]
 pub struct AssistantsApiResponseFormat {
-    /// Must be one of `text` or `json_object`.
+    /// Must be one of `text`, `json_object` or `json_schema`.
     pub r#type: AssistantsApiResponseFormatType,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub json_schema: Option<serde_json::Value>,
```
nytopop, line 136: Would this be better as a struct? Like the functions api, the actual json schema is a field inside of an object with defined fields in the docs:
2024-08-12-205225_389x547_scrot

nytopop (approved):
LGTM pending a rustfmt but I'm not a maintainer, so :shrug:

64bit:
@domenkozar PR is always appreciated.

However it required a bit more work for following issues, which addressed  in #257 :
1. While the change may work, they are semantically inaccurate, for example it cant prevent you to construct an object type type text and still having json_schema. 
2.  In addition, it may also lead to  issue when an object is serialized and deserialized and they end up not being same (please see https://github.com/64bit/async-openai/issues/216)
3. No supporting tests please see https://github.com/64bit/async-openai?tab=readme-ov-file#contributing 

Hence, this PR can be closed. Thank you again!

PR #263, opened 2024-09-02
Fix service_tier serialization for CreateChatCompletionRequest

tomgie:
As per the API reference page on OpenAI's website, the parameter can be excluded from the request.
https://platform.openai.com/docs/api-reference/chat


64bit (approved):
Thank you for the fix!

PR #267, opened 2024-09-10
Fix content, refusal serialization for CreateChatCompletionRequest

chirino:
They are optional per the openai docs.


64bit (approved):
Thank you for the fix!

PR #274, opened 2024-10-05
allow json values as function call arguments

linusbierhoff:
Currently, only strings (which can be in JSON format) are allowed.

Models like LLAMA, deployed with the Hugging Face Text Generation API, are accessible with the same API as OpenAI, but generate a JSON object instead of a string.


linusbierhoff:
Changed from enum to function, which always deserialize in string

64bit:
Thank you for the PR, it is always welcome!

I'd recommend to add this in community crates dedicated to LLAMA instead, unfortunately for async-openai this is out of scope. For more information please see https://github.com/64bit/async-openai?tab=readme-ov-file#contributing

linusbierhoff:
@64bit, thank you for your kind reply.
I understand that it's out of scope. Anyway, since it's only a small change that does not affect the OpenAI compatibility, it would be perfect if we could merge it. I think many developers use the [langchain-rust crate](https://github.com/Abraxas-365/langchain-rust), which relies on this package, and the parsing issue of the function calls makes it difficult to use.

64bit:
There's an idea on how to support it while keeping the scope limited to OpenAI: https://github.com/64bit/async-openai/issues/280#issuecomment-2408744933 

Please consider it for contributions

64bit:

PR #284, opened 2024-10-18
make prompt_tokens_details and completion_tokens_details optional

e-max:
Legacy models such as GPT3.5 doesn't return these fields, so we make them optional.


64bit:
It seems that spec was even updated even more, its been addressed in v0.26.0. Thank you for your prompt response though!

PR #287, opened 2024-11-13
Generate extensible versions of methods by allowing generic requests and response

### LINKED ISSUE #280: Support Azure content filters in completion
Hello,

we currently use `async-openai` to query Azure OpenAI and we would need support for the new [content filters](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter) on Azure OpenAI. Basically it's a new `content_filter_results` field in `(Chat)Choice`.

I would be happy to work on it, however I read the `Contributing` section of the README :smile:  and I understand you want to stick to OpenAI specs, while this is an Azure extension. But also (thankfully for us) there are already some Azure specific stuff I reckon (e.g. the AzureConfig), although limited admittedly.

What would be the best course of action? I see a few options:

1. add the support, but make this field optional (it is anyway since you can choose to not enable those filters) so it doesn't break on pure OpenAI
2. add the support and add an `azure` feature to the crate and gate Azure-specific stuff behind it
3. add some sort of "extra fields" to `(Chat)Choice` where anything extra would be stored as a `serde_value::Map` (and then people "do their thing" with those extra fields) so it supports "extra features" from providers without being Azure-specific
4. I can also just fork (a friendly fork! or even just an internal one)

`1` would work for me of course, but I think `2` (my preference) would be interesting, to have both pure OpenAI as default, but also support Azure extensions, since I believe Azure is actually a big de facto OpenAI provider, but I understand if you're not interested / willing to. `3` sounds quite hackish. `4` is OK of course as last resort. What do you think?

Sufflope:
Fixes #280

### DIFF
--- a/async-openai/src/types/mod.rs
+++ b/async-openai/src/types/mod.rs
@@ -72,3 +72,8 @@ impl From<UninitializedFieldError> for OpenAIError {
         OpenAIError::InvalidArgument(value.to_string())
     }
 }
+
+pub trait RequestForStream {
+    fn is_request_for_stream(&self) -> bool;
+    fn set_request_for_stream(&mut self, stream: bool);
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = [ "async-openai", "examples/*" ]
+members = [ "async-openai", "async-openai-macros", "examples/*" ]
 # Only check / build main crates by default (check all with `--workspace`)
-default-members = ["async-openai"]
+default-members = ["async-openai", "async-openai-macros"]
 resolver = "2"
--- a/async-openai-macros/Cargo.toml
+++ b/async-openai-macros/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "async-openai-macros"
+version = "0.1.0"
+authors = ["Jean-Sébastien Bour <jean-sebastien@bour.name>"]
+categories = ["api-bindings", "web-programming", "asynchronous"]
+keywords = ["openai", "async", "openapi", "ai"]
+description = "Procedural macros for async-openai"
+edition = "2021"
+rust-version = "1.70"
+license = "MIT"
+readme = "README.md"
+homepage = "https://github.com/64bit/async-openai"
+repository = "https://github.com/64bit/async-openai"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+darling = "0.20"
+itertools = "0.13"
+proc-macro2 = "1"
+quote = "1"
+syn = "2"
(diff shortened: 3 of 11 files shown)


Sufflope:
For the moment I only used the macro on my use case, would you prefer to add it to all relevant methods in the code, or keep it "on demand" when someone needs an extension for some use case?

I guess we should document this mechanism somewhere in the readme too.

Also the `impl Streamable` could easily be a derive macro.

async-openai/src/types/mod.rs
```diff
@@ -72,3 +72,7 @@ impl From<UninitializedFieldError> for OpenAIError {
         OpenAIError::InvalidArgument(value.to_string())
     }
 }
+
+pub trait Streamable {
```
64bit, line 76: Seems like this will become public API and not just internal to crate, in which case `Streamable` is very overloaded word and open to individual interpretation, becauses usually responses are "streamable" and not requests.

I'd suggest renaming it. Some suggestions like: `trait RequestForStream` or method to be `is_request_for_stream` or something more appropriate where its less open to interpretation.

64bit:
> For the moment I only used the macro on my use case, would you prefer to add it to all relevant methods in the code, or keep it "on demand" when someone needs an extension for some use case?

I checked both [gemini](https://ai.google.dev/gemini-api/docs/openai) and [ollama](https://ollama.com/blog/openai-compatibility), so for practical purposes doesn't require to be added on all APIs and we can resort to on-demand additions (must be documented though - more below).

> 
> I guess we should document this mechanism somewhere in the readme too.

Absolutely, please document it on both README.md so Github repo and crates.io visitor knows they can "Bring your own types" and also in `lib.rs` so that visitors of docs.rs learn about it too.

My primary motivation to use this feature is that I can escape deserialization errors and just plugin `serde_json::Value` when I'm experimenting/working with compatible API providers. 

> 
> Also the `impl Streamable` could easily be a derive macro.

Its one liner code, not a huge burden, hence not required.

64bit:
Also please do add [test](https://github.com/64bit/async-openai/tree/main/async-openai/tests), it can be very minimal to just compile on a custom request and response.

Sufflope:
Thanks for the review, I should have addressed all remarks. Writing good documentation is hard :sweat_smile: Let me know what you think.

Sufflope:
Hi, a small ping to know if you had time to check the new version with changes wrt. your comments?

64bit:
Thank you for your contributions! 

Please consider using [Bring your own types](https://github.com/64bit/async-openai?tab=readme-ov-file#bring-your-own-types) feature in v0.28.0

PR #288, opened 2024-11-16
Fix chunking strategy serialization and field visibility

nikessel:
# Fix chunking strategy serialization and field visibility

## Description
This PR fixes three issues with chunking strategies in the vector stores API:

1. Incorrect serialization format for static chunking strategy
2. Incorrect casing in enum variant serialization ("Static" vs "static")
3. Private fields preventing users from setting chunking parameters

### Serialization Fix
The OpenAI API expects the static chunking configuration to be nested under a "static" field, but the current implementation serializes it directly. This causes the error:
```
"Missing required parameter: 'chunking_strategy.static'."
```

Additionally, the enum variants were serializing with incorrect casing, causing:
```
"Invalid value: 'Static'. Supported values are: 'auto' and 'static'."
```

Changed from:
```rust
#[serde(tag = "type")]
pub enum VectorStoreChunkingStrategy {
    #[default]
    Auto,
    Static(StaticChunkingStrategy),
}
```

To:
```rust
#[serde(tag = "type")]
pub enum VectorStoreChunkingStrategy {
    #[default]
    #[serde(rename = "auto")]
    Auto,
    #[serde(rename = "static")]
    Static {
        #[serde(rename = "static")]
        config: StaticChunkingStrategy,
    },
}
```

### Field Visibility
Made fields in `StaticChunkingStrategy` public to allow users to configure chunking parameters:

```rust
pub struct StaticChunkingStrategy {
    pub max_chunk_size_tokens: u16,
    pub chunk_overlap_tokens: u16,
}
```

### DIFF
--- a/async-openai/src/types/assistant.rs
+++ b/async-openai/src/types/assistant.rs
@@ -63,7 +63,10 @@ pub enum AssistantVectorStoreChunkingStrategy {
     #[serde(rename = "auto")]
     Auto,
     #[serde(rename = "static")]
-    Static(StaticChunkingStrategy),
+    Static {
+        #[serde(rename = "static")]
+        config: StaticChunkingStrategy,
+    },
 }
 
 /// Static Chunking Strategy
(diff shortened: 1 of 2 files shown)


64bit (approved):
Thank you for detailed description and PR!

PR #289, opened 2024-11-18
Rename max output tokens as per api docs

codesoda:
As per the docs at https://platform.openai.com/docs/api-reference/realtime-client-events/session/update

The name for MaxOutputTokens is `max_response_output_tokens`. I've renamed the enum and attribute accordingly.


64bit (approved):
Thank you for the fix!

PR #290, opened 2024-11-23
chore(deps)!: Update Secrecy

timonv:
Updates Secrecy to the latest version. Because the api changed, this is unfortunately a breaking change (swiftide included 🤷).

However, probably best to keep secret managing libraries up to date.

Thanks for the great library, we use it *a lot*.


64bit:
Glad to learn its useful for you, Thank you for the PR! 

There are few breaking changes in main already which are just updated types, those changes are easy for users to upgrade as compared to the changes in this PR - hence I'll plan to release it in subsequent releases (and  take opportunity to update all other dependencies as well) - please consider release in v0.28.0+ for this PR.

PR #296, opened 2024-12-08
chat/completions api suport vioce input/output

laonger:
chat/completions api suport vioce input/output

* CreateChatCompletionRequest updated

* CreateChatCompletionResponse updated

* add example: /example/chat-audio


64bit:
Thank you for PR! +1 for including example.

On an initial scan, it looks like some of the types may already exist in the repo, do you mind double checking that the types are not defined twice?

laonger:
> Thank you for PR! +1 for including example.
> 
> On an initial scan, it looks like some of the types may already exist in the repo, do you mind double checking that the types are not defined twice?

Hi, thank you for your reply.

I have checked the whole project, i did not find any data structure that exactly matches my code.

Would you please point them out for me?

Thanks very much.

async-openai/src/types/chat.rs
```diff
@@ -492,6 +545,38 @@ pub enum ServiceTierResponse {
     Default,
 }
 
+#[derive(Clone, Serialize, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum ChatCompletionModalitie {
```
64bit, line 550:
```suggestion
pub enum ChatCompletionModalities {
```

64bit:
There are bunch of refactors that requires here, names are not matching spec. Also why is there a builder pattern on a response object?  Also example contain type which is not present?

For this reason I cannot accept this PR as it is.

64bit:
My initial comment was because types are confusing between Chat and Audio APIS.

It's addressed in v0.27.0. 

For this PR you're welcome to update the example, I'll happily accept the example. For mp3 file for input you can copy it from other speech to text example in the repo.

Issue #298, opened 2024-12-12
Add store parameter to request


hugohmn:
Hello,

Would be great to add the new parameter **store** to requests, to be able to use evaluations and distillation features.

see :
https://community.openai.com/t/chat-completions-history-on-openai-dashboard/987804
https://platform.openai.com/docs/guides/distillation/send-fine-tuned

Maybe something that would look like this :

```rust
let request = CreateChatCompletionRequestArgs::default()
        .max_tokens(512u32)
        .model("gpt-4o-mini")
        .messages([
            ChatCompletionRequestUserMessageArgs::default()
                .content("Hi")
                .build()?
                .into(),
        ])
        .store(true)
        .build()?;
```

Thanks

64bit:
This is available in v0.27.0

PR #300, opened 2024-12-17
Add store and metadata parameters

didriksg:
### Description
Solves #298 by adding support for store flag and metadata JSON for tags in chat

### Changes
- CreateChatCompletionRequest updated
- Add example: /example/chat-store 
  - Match examples in https://platform.openai.com/docs/guides/distillation and https://platform.openai.com/docs/guides/evals


didriksg:
Chat completions in openai platform dashboard after running the provided example:

image

PR #307, opened 2024-12-29
Fix comment formatting

ysimonson:


64bit:
Thanks! it was fixed by spec sync in recent releases

PR #309, opened 2024-12-30
Fix parsing of inf strings

ysimonson:
Server events `MaxResponseOutputTokens` will currently not parse:

```
called `Result::unwrap()` on an `Err` value: JSON decode error: data did not match any variant of untagged enum MaxResponseOutputTokens
```

This fixes the issue, though it's a hack as it doesn't validate that the string is `inf`. I'm not sure how to do that in serde without writing a custom deserializer/visitor.


64bit:
Thanks! it was fixed by spec sync in recent releases

PR #310, opened 2024-12-30
Fix parsing of cancelled responses

ysimonson:
Without this, cancelled responses fail to parse:

```
called `Result::unwrap()` on an `Err` value: JSON decode error: unknown variant `cancelled`, expected `incomplete` or `failed`
```


64bit:
Fixed in recent releases

Issue #271, opened 2024-09-27
[Feature Request] Gemini support


Napolitain:
i know it is a bit "out of scope" seemingly but I think we would benefit from Gemini support using the same interface.
main api could stay identical to openai and thus not implement specifics from google
there is also some ai audio stuff in google ai (non gemini)

64bit:
Nobody stopping you from doing so, pretty liberal license to fork an do it.

dashed:
You can use the Gemini models now: https://developers.googleblog.com/en/gemini-is-now-accessible-from-the-openai-library/

PR #311, opened 2025-01-09
Make id field in the completion response optional

quangIO:
This allows to use async-open library for Google gemini (https://ai.google.dev/gemini-api/docs/openai) as well. 

ref: #271


64bit:

PR #317, opened 2025-01-12
remove async-convert and bump rust-version

### LINKED ISSUE #316: Remove async-convert
As in [my commit](https://github.com/ifsheldon/async-openai-wasm/commit/d82a60563d4227ba93d6920016624e9915e6737b) to the downstream fork `async-openai-wasm`, it's possible to remove async-convert crate. Without async-convert, we don't need to use `async-trait` either, which remove extra boxing. This also shrinks binary size. I would be happy to upstream this change in a PR.

ifsheldon:
This fixes #316 

As per https://github.com/rust-lang/api-guidelines/discussions/231, the increase of MSRV is not a breaking change, so there's no need for a major release.


64bit (approved):
Thank you for the PR!

PR #318, opened 2025-01-13
Add `From<ChatCompletionRequestDeveloperMessage>` implementation for `ChatCompletionRequestMessage`

keithschulze:
In trying to use the new `ChatCompletionRequestDeveloperMessage` added in #312, I found I couldn't use `.into()` to add it to `messages` because a `From<ChatCompletionRequestDeveloperMessage>` implementation for `ChatCompletionRequestMessage` wasn't available.

### DIFF
--- a/async-openai/src/types/impls.rs
+++ b/async-openai/src/types/impls.rs
@@ -553,6 +553,12 @@ impl From<ChatCompletionRequestSystemMessage> for ChatCompletionRequestMessage {
     }
 }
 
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you for both contributions! 

I'll happily accept both, for #319  because its adding new dependency I'd appreciate if we use the value it provides more, there are a lot of Enums in the crate and seems like `derive_more::From` can be applied to all of them.

PR #319, opened 2025-01-13
Add missing From impl for ChatCompletionRequestMessage, replace trivial From impls with derive_more

grantslatton:
`ChatCompletionRequestMessage` didn't have `From<ChatCompletionRequestDeveloperMessage>`

I went to add it, and found the TODO:

```
// todo: write macro for bunch of same looking From trait implementations below
```

The crate `derive_more` does this by deriving `From` for simple wrapper-branching enums like this.


64bit:
Note from previous PR: https://github.com/64bit/async-openai/pull/318#pullrequestreview-2560514753

Please consider using `derive_more::From` for all Enums for consistency.

PR #326, opened 2025-01-28
✨Support new voices

### LINKED ISSUE #321: Support new voices
https://platform.openai.com/docs/guides/text-to-speech

The OpenAI docs list support for nine voices, however, the enums in this library only support a subset of them. 

https://github.com/64bit/async-openai/blob/c48e62e72b1c8bd9658af7e682c996ba5c93e083/async-openai/src/types/audio.rs#L39-L47

https://github.com/64bit/async-openai/blob/c48e62e72b1c8bd9658af7e682c996ba5c93e083/async-openai/src/types/chat.rs#L613-L622

I'm not sure why the voices in our `chat.rs` and `audio.rs` file are already different, but is it possible to add all the voices listed in the OpenAI docs to this API so that we can generate audio for all the voices? I'm doing text-to-speech conversion using the enum in the audio.rs file and I'd like the newer voice options.

sloganking:
Fixed #321

I tested this locally on the most up-to-date version of the repository `0.27.2` and it works for me.

### DIFF
--- a/async-openai/src/types/audio.rs
+++ b/async-openai/src/types/audio.rs
@@ -32,17 +32,19 @@ pub enum SpeechResponseFormat {
     Pcm,
     Wav,
 }
-
 #[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq)]
 #[serde(rename_all = "lowercase")]
 #[non_exhaustive]
 pub enum Voice {
     #[default]
     Alloy,
+    Ash,
+    Coral,
     Echo,
     Fable,
     Onyx,
     Nova,
+    Sage,
     Shimmer,
 }
 


sloganking:
@64bit Any thoughts on this? It's a small pull request and should be good to go.

64bit:
Please fix formatting

sloganking:
Hey @64bit I see this was closed rather than merged. 

Is it because I didn't update the formatting, like you asked? Or because you don't want this feature, or got it implemented a different way? 

Is the formatting you're referring to the deleted line on top of the enum declaration or something else?

sloganking:
@64bit Any update here?

PR #329, opened 2025-02-04
Fixed duplicated arguments in tool-call-stream example

mmende:
When playing around with the tool-call-stream example in combination with [mistral's API](https://console.mistral.ai/) I encountered a strange bug where the arguments where duplicated (e.g. instead of `{ "city": "Berlin" }` I would always get `{ "city": "Berlin" }{ "city": "Berlin" }`).

This issue is caused by mistral's API streaming the arguments together with the tool call id and function name which in the example results in the `ChatCompletionMessageToolCall` being inserted including arguments and then the arguments would be concatenated to itself again.

The adjusted example in this PR fixes this issue.


b0xtch:
Yeah noticed that same with groq mistral model as well

64bit:
Thanks, The example was updated hence closing this

PR #334, opened 2025-02-09
add reasoning_content support for deepseek r1

lijingrs:
add reasoning_content support for deepseek r1


64bit:

Issue #173, opened 2024-01-05
Changing base url - Usage with open source LLM - Invalid status code: 404 Not Found


louis030195:
Hey I'm trying to use `async-openai` with `axum` and open source LLMs through `perplexity.ai` in my test.

Basically my endpoint would route the request to OpenAI API or an OpenAI API like API changing the URL of the API based on the given model like `gpt-4` would go to openai and `mistralai/whatever` would go to `MODEL_URL`

Getting a 404. Not sure if I'm doing something wrong or this use case is implemented?

These are my env var
```
MODEL_API_KEY="pplx-..."
MODEL_URL="https://api.perplexity.ai/chat/completions"
```

My code:

```rs
use async_openai::Client;
use async_openai::{config::OpenAIConfig, types::CreateChatCompletionRequest};
use axum::{
    extract::{Extension, Json, Path, State},
    http::StatusCode,
    response::IntoResponse,
    response::Json as JsonResponse,
};

use async_stream::try_stream;
use axum::response::sse::{Event, KeepAlive, Sse};
use futures::Stream;
use futures::StreamExt;
use serde::{Deserialize, Serialize};
use std::convert::Infallible;
use std::error::Error;
use std::io::{stdout, Write};
use tokio::sync::broadcast::Receiver;
use tokio_stream::wrappers::BroadcastStream;
use url::Url;

fn extract_base_url(model_url: &str) -> Result<String, url::ParseError> {
    let url = Url::parse(model_url)?;
    let base_url = url.join("/")?;
    Ok(base_url.as_str().to_string())
}
// copied from https://github.com/tokio-rs/axum/discussions/1670

pub async fn stream_chat_handler(
    Json(request): Json<CreateChatCompletionRequest>,
) -> Result<Sse<impl Stream<Item = Result<Event, Infallible>>>, (StatusCode, String)> {
    let model_name = &request.model;
    let model_url = std::env::var("MODEL_URL")
        .unwrap_or_else(|_| String::from("http://localhost:8000/v1/chat/completions"));
    let base_url = extract_base_url(&model_url).unwrap_or_else(|_| model_url);
    let (api_key, base_url) = if model_name.contains("/") {
        // Open Source model
        (std::env::var("MODEL_API_KEY").unwrap_or_default(), base_url)
    } else {
        // OpenAI model
        (
            std::env::var("OPENAI_API_KEY").unwrap_or_default(),
            String::from("https://api.openai.com"),
        )
    };
    let client = Client::with_config(
[...]

PR #336, opened 2025-02-17
Add simple ollama chat example

### LINKED ISSUE #191: Ollama AI Chat Completion Example / OpenAI Compatibility
Given that officially Chat completion API is compatibale - this library should work without any changes. Hence it would be great to have a fully self contained example which works with local ollama using this crate

https://ollama.com/blog/openai-compatibility

matthewhaynesonline:
Fixes #191. 

I used the code examples from https://github.com/64bit/async-openai/issues/173 as a starting point and tried to simplify them to match the existing chat example.

### DIFF
--- a/examples/ollama-chat/README.md
+++ b/examples/ollama-chat/README.md
@@ -1,0 +1,38 @@
+## Setup
+
+A docker compose file is provided to run a dockerized version of Ollama and download a default model. You will need the Ollama container to be up and running _before_ you can run the Rust example code.
+
+You can check the container status with `docker ps` or check the container's logs with `docker container logs {CONTAINER NAME} -f`. E.g. `docker container logs ollama -f`.
+
+## Running the Example
+
+```sh
+# Bring ollama up with model and wait for it to be healthy.
+docker compose up -d
+
+# Once model is downloaded and Ollama is up, run the Rust code.
+cargo run
+```
+
+## Docker Notes
+
+- Since Ollama requires you to pull a model before first use, a custom entrypoint script is used. See [Stack Overflow discussion](https://stackoverflow.com/a/78501628).
+  - The model will be cached in the volumes dir.
+    - Depending on your network connection, the healthcheck may need to be adjusted to allow more time for the model to download.
+- [llama3.2:1b](https://ollama.com/library/llama3.2:1b) is used in the example as it is a smaller model and will download more quickly compared to larger models.
+  - A larger model will provide better responses, but be slower to download.
+  - Also, using the default CPU inference, smaller models will have better tokens / second performance.
+- The GPU mapping is written but commented out. This means it will default to CPU inference which is slower, but should run without any additional setup.
+  - If you have a GPU and the proper container support, feel free to uncomment / adapt.
+
+## Ollama OpenAI Compatibility
+
+**NOTE: an api key parameter is used for compatibility with OpenAI's API spec, but it is ignored by Ollama (it can be any value).**
+
+See the [Ollama OpenAI Compatibility docs](https://github.com/ollama/ollama/blob/main/docs/openai.md) for more details on what Ollama supports.
+
+## Response
+
+> Response:
+>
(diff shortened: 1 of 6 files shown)


async-openai/README.md
```diff
@@ -59,6 +59,11 @@ $Env:OPENAI_API_KEY='sk-...'
 - Visit [examples](https://github.com/64bit/async-openai/tree/main/examples) directory on how to use `async-openai`.
 - Visit [docs.rs/async-openai](https://docs.rs/async-openai) for docs.
 
+### Local LLM Usage
```
64bit, line 62: This is not required. Ollama example has its own README

examples/ollama-chat/README.md
```diff
@@ -0,0 +1,18 @@
+## Prerequisites
```
64bit, line 1: Please consider adding fully contained example - in this case if there's dependency on ollama - please consider creating docker-compose file to bring it up in single command. 

I'd prefer to have quick way to run an example, instead having them additional steps - which a user may already do on their own

matthewhaynesonline:
Hi, I've updated my PR with your feedback, though the Docker setup has some caveats I didn't have elegant solutions for. 

1. **Build context:** since Docker restricts where you can copy files from to the current directory branch (parents / siblings are not allowed), I adjusted the example to pull the `async_openai` crate from crates.io, not the parent dir. 
   1. The only way I could think to use the crate from the parent project dir was to place the `docker-compose.yml` / `Dockerfile` in the project root, but that would no longer make the example self contained.
2. **First time Ollama model pull:** Ollama requires you to pull a model before you can use it and, understandably, the Ollama docker image doesn't come preloaded with any models. As such, a custom entrypoint is use to pull the model on startup. A healthcheck is then used to wait to start the Rust container until after the Ollama container is fully ready.

What do you think?

64bit:
Thanks for update, I just tried it and looks good! 

I think `Dockerfile` and `ollam_chat` containers  in docker-compose are not required. My initial reason for docker-compose was to bring up Ollama and then `cargo run` example which would points to "localhost:<ollama-port>".

Your entrypoint stuff to download model looks good to me.

So user has to do two steps:
```
# bring ollama up with model
docker compose up -d

# once model is downloaded and ollama is up
cargo run 
```

examples/ollama-chat/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rust:1.85 AS builder
```
64bit, line 1: Not required

examples/ollama-chat/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    command: ollama-chat
+    restart: on-failure:3
+    environment:
+      API_BASE: "http://ollama:11434/v1"
+      API_KEY: "ollama"
+      MODEL: "llama3.2:1b"
+    depends_on:
+      ollama:
+        condition: service_healthy
+
```
64bit, line 14: not required

examples/ollama-chat/ollama_entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+# Start Ollama in the background.
+/bin/ollama serve &
+# Record Process ID.
+pid=$!
+
+# Pause for Ollama to start.
+sleep 5
+
+echo "Retrieving model $MODEL..."
+ollama pull $MODEL
```
64bit, line 12: Nice

examples/ollama-chat/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    types::{
+        ChatCompletionRequestAssistantMessageArgs, ChatCompletionRequestSystemMessageArgs,
+        ChatCompletionRequestUserMessageArgs, CreateChatCompletionRequestArgs,
+    },
+    Client,
+};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let api_base = env::var("API_BASE").expect("API_BASE is not set in the environment.");
```
64bit, line 14: This can be hard coded to what docker compose will bring up 

so `localhost:11434`

matthewhaynesonline:
@64bit Sounds good! I updated my branch with your feedback to run the Rust code natively and rebased / cleaned up the commit history off latest upstream main.

64bit (approved):
Looking good, thank your for your contributions!

PR #337, opened 2025-02-19
Make ID optional in ChatCompletionResponse, Gemini doesn't have it

jtwaleson:
I was trying to get Google Gemini integrated and their openai compatible endpoint doesn't include an ID field. I've tested this change and it works.

Not sure if you are open to a PR like this as the ID is present in the OpenAI OpenAPI spec https://platform.openai.com/docs/api-reference/chat/get, and it is technically an issue with the Google Gemini OpenAI compatibility layer.


Neboer:
This is very helpful, the library cannot support Gemini OpenAI API unless merge this change.

64bit:

Issue #340, opened 2025-02-28
OpenAI Updated Assistants Schema - Small Fix


hiibolt:
It seems that with the launch of GPT4.5 they made a minor change to some return types:

```
Error: Failed to retrieve assistant

Caused by:
    failed to deserialize api response: unknown variant `default-2024-08-21`, expected `auto` or `default_2024_08_21` at line 19 column 3
```

It looks related to this [struct](https://github.com/64bit/async-openai/blob/c48e62e72b1c8bd9658af7e682c996ba5c93e083/async-openai/src/types/assistant.rs#L150), with the only change being using hyphens instead of underscores.

hiibolt:
Minimal reproducible:
```rust
#[tokio::main]
async fn main() -> Result<()> {
    let client = Client::new();

    // Initialize the assistant
    let assistant = client
        .assistants()
        .retrieve("asst_...")
        .await
        .context("Failed to retrieve assistant")?; // Panics here

    Ok(())
}
```

irbull:
+1, I'm hitting this too. However, their docs still mention `default_2024_08_21`.

https://platform.openai.com/docs/assistants/tools/file-search

> ranker - Which ranker to use in determining which chunks to use. The available values are auto, which uses the latest available ranker, and default_2024_08_21.

There have been a number of problems with OpenAI in the past few days with the 4.5 release.

PR #341, opened 2025-02-28
bugfix: Updated `FileSearchRanker` to match change by OpenAI

hiibolt:
Fixes issue mentioned in #340

### DIFF
--- a/async-openai/src/types/assistant.rs
+++ b/async-openai/src/types/assistant.rs
@@ -147,7 +147,7 @@ pub struct AssistantToolsFileSearchOverrides {
 pub enum FileSearchRanker {
     #[serde(rename = "auto")]
     Auto,
-    #[serde(rename = "default_2024_08_21")]
+    #[serde(rename = "default-2024-08-21")]
     Default2024_08_21,
 }
 


irbull:
I can confirm that if I use `curl` to access the create run endpoint, it also returns `default-2024-08-21` which appears to be different from what the docs say. Since this appears to be an undocumented change, what do you think about using an `alias` instead of a rename, and supporting both `default-2024-08-21` and `default_2024_08_21`. Just in case they _break_ their API again.

Something like:
```
pub enum FileSearchRanker {
    #[serde(rename = "auto")]
    Auto,
    #[serde(alias = "default_2024_08_21")]
    #[serde(alias = "default-2024-08-21")]
    Default2024_08_21,
}
```

hiibolt:
> I can confirm that if I use `curl` to access the create run endpoint, it also returns `default-2024-08-21` which appears to be different from what the docs say. Since this appears to be an undocumented change, what do you think about using an `alias` instead of a rename, and supporting both `default-2024-08-21` and `default_2024_08_21`. Just in case they _break_ their API again.
> 
> Something like:
> 
> ```
> pub enum FileSearchRanker {
>     #[serde(rename = "auto")]
>     Auto,
>     #[serde(alias = "default_2024_08_21")]
>     #[serde(alias = "default-2024-08-21")]
>     Default2024_08_21,
> }
> ```

I agree, I will link a commit tonight.

hiibolt:
@64bit Could you look this change over? Because of the unprompted and undocumented spec change by OpenAI, crates currently depending on using the Assistants API fail even if they do not directly use this struct

PR #347, opened 2025-03-12
fix: readme example link

attila-lin:


64bit (approved):
Thank you for the fix!

Issue #42, opened 2023-03-01
Add ChatGPT API


m1guelpf:
See https://platform.openai.com/docs/api-reference/chat/create

PR #353, opened 2025-03-26
feat: Gemini openai compatibility

DarshanVanol:
* Added examples demonstrating Gemini API integration with async_openai Rust library.


DarshanVanol:
Closes #42

64bit (changes requested):
Thank you for PR. Please take a look at CONTRIBUTION guidelines and Issue

> This library should work as is if Gemini "OpenAI compatibility" is fully compatible with OpenAI. Otherwise use byot feature on need-basis.

64bit (approved):
Thanks for this, great work with documentation and covering all APIs!

Happy to merge this, can you please drop the links/references to copyrighted contents in sample_data  (image and audio) used in this PR?  

Either one of : having link in README (preferred) of the example or here in comments would be OK.

PR #354, opened 2025-04-03
Backoff when OpenAI returns 5xx

tinco:
Hi! Thanks for creating and maintaining async-openai. I noticed during the outtages last weekend that the client is not dealing with server errors as is instructed by OpenAI.

I'm fairly certain these are new instructions, on this page you can see:

https://platform.openai.com/docs/guides/error-codes

That they are now advising to retry (after a delay) on any server error.

It is slightly more involved than that, because I also noticed that it is sometimes (always?) the case that they do not return JSON on server errors, so instead of parsing JSON, we need to create the `ApiError` ourselves by simply inserting the whole response body into the message.

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -8,7 +8,7 @@ use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{
     config::{Config, OpenAIConfig},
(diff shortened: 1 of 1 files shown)


64bit (approved):
Nice full context in your description, appreciate it. 

Thank you for the PR!

PR #357, opened 2025-04-22
add support for streaming web annotations

ehayes2000:
Add support for web-annotation streaming.


async-openai/src/types/chat.rs
```diff
@@ -940,6 +940,27 @@ pub struct ChatCompletionStreamResponseDelta {
...
+    /// The url source annotations provided by search
+    pub annotations: Option<Vec<Annotation>>,
+}
+
+#[derive(Deserialize, Debug, Serialize, Clone, PartialEq)]
+#[serde(untagged)]
+pub enum Annotation {
+    UrlCitation {
+        #[serde(rename = "type")]
+        c_type: String,
```
cyborg42, line 952: can use r#type instead

PR #359, opened 2025-04-24
feat: tool management

cyborg42:
add tool management for chat


cyborg42:
I've added Responses API compatibility, but I can only test it with OpenRouter, which doesn't support the Responses API. Could you test it using the new responses-function-call example? Also, I haven't implemented Responses API for `ToolCallStreamManager`. Should we create a new dev branch to complete this?

cyborg42:
I noticed that the Responses API currently doesn't support streaming, so we can focus on `ToolCallStreamManager` in the future.

64bit:
Response API streaming support was added in https://github.com/64bit/async-openai/pull/405

PR #360, opened 2025-04-28
Implement vector store search, retrieve file content operations

cfraz89:
- Implement vector store search as in
  - [File Search guide](https://platform.openai.com/docs/guides/retrieval ) 
  - [Search api](https://platform.openai.com/docs/api-reference/vector-stores/search) 
  - [OpenAPI spec](https://github.com/openai/openai-openapi/blob/76489c0d59a688c2116307e9941fd01f6caacb94/openapi.yaml#L14444)

A basic query looks like this:
```rust
    let results = client
        .vector_stores()
        .search(
            &vector_store.id,
            VectorStoreSearchRequest {
                query: "uber profit".into(),
                ..Default::default()
            },
        )
        .await?;
```

- Also implement retrieve file contents api
- Add attributes property to CreateVectorStoreFileRequest
- Add an example for vector store search.


examples/vector-store-retrieval/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+            VectorStoreSearchRequest {
+                query: "uber profit".into(),
+                ..Default::default()
+            },
+        )
+        .await?;
+
+    // Print the search results
+    println!("Search results: {:#?}", results);
+    Ok(())
```
64bit, line 71:
```suggestion
    // Cleanup to avoid costs 
    let _ = client
        .vector_stores()
        .delete(&vector_store.id)
        .await?;
    
    let _ = client
        .files()
        .delete(&uber_file.id)
        .await?;
    
    let _ = client
        .files()
        .delete(&lyft_file)
        .await?;
    Ok(())
```

examples/vector-store-retrieval/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        .await?;
+    
+    let _ = client
+        .files()
+        .delete(&uber_file.id)
+        .await?;
+    
+    let _ = client
+        .files()
+        .delete(&lyft_file)
```
64bit, line 84:
```suggestion
        .delete(&lyft_file.id)
```

64bit (approved):
Thank you for the PR, really quality work - from matching spec to adding example! 

I committed a few suggestions to cleanup after example is run - delete vector store and files to avoid any costs.

Issue #363, opened 2025-05-08
CreateImageEditRequest only deals with a single image


spantaleev:
With most models, the `/v1/images/edits` API only supports a single image in the `image` field.

However, with the new `gpt-image-1` model, multiple images may be provided.

See https://platform.openai.com/docs/api-reference/images/createEdit#images-createedit-image 

> For `gpt-image-1`, each image should be a png, webp, or jpg file less than 25MB. You can provide up to 16 images.
> 
> For `dall-e-2`, you can only provide one image, and it should be a square png file less than 4MB.

PR #366, opened 2025-05-10
Add support for multiple image files in `CreateImageEditRequest`

spantaleev:
Fixes https://github.com/64bit/async-openai/issues/363


64bit:
Thank you for the PR! 

This can be improved to match the spec `image string or array Required`
https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L21129-L21137

That's because a user can input a single or multiple images in builder like:

**single image**
```
  CreateImageEditRequestArgs::default()
     .image(single_image)
     ...
    .build()?
```

**multiple image**
 
```
  CreateImageEditRequestArgs::default()
     .image([image1, image2])  # this can be vector, or just [] 
     ...
    .build()?
```

There are pattern in the repo which makes it happen in other places, and then implement `From` traits to have API like above.

64bit (approved):
Previous comment is good to have - can be added later.

Thank you for the PR!

PR #369, opened 2025-05-18
Expose a way to customize the path to the chat completions endpoint

Kerollmops:
Some APIs are close to being compatible with this crate. The byot (bring your own type) feature is handy, but useless when the path is incompatible.

For example, [Ollama's chat completions route](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion) is close to the OpenAI API but reachable on the `/api/chat` route. On the other hand, Claude (Anthropic) is available through the [`/v1/messages` route](https://docs.anthropic.com/en/api/messages), which is also very close to being compatible with the OpenAI API.

### DIFF
--- a/async-openai/src/chat.rs
+++ b/async-openai/src/chat.rs
@@ -12,11 +12,21 @@ use crate::{
 /// Related guide: [Chat completions](https://platform.openai.com//docs/guides/text-generation)
 pub struct Chat<'c, C: Config> {
     client: &'c Client<C>,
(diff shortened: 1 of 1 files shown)


64bit:
The latest release in crate now has support for configuring path, query or headers for an individual request.

More here: https://github.com/64bit/async-openai?tab=readme-ov-file#configurable-request

(hence closing this PR)

PR #373, opened 2025-05-21
feat: Add responses API

samvrlewis:
Adds support for the OpenAI responses API.

Doesn't have support for streaming yet.

~Due to types being reexported from the root of the crate, there's a few types with existing names (but different shapes) that I've exported with `Responses` prefixes to avoid making ambiguous types. For example:  `ResponsesFilePath`,  `ResponsesRole`). Happy for feedback if there's a different way that this would be preferred.~ (edit: realised later it would probably be cleaner just to not export all the types at the root, as there's too much duplication).


twitchax:
Cool.

I tried this against an existing project, and I think something might be wrong with the `text()` option.  I get.

```
Error while handling: invalid_request_error: Missing required parameter: 'text.format.name'. (param: text.format.name) (code: missing_required_parameter)
```

I am guessing that in chat.rs,

```rust
pub enum ResponseFormat {
    /// The type of response format being defined: `text`
    Text,
    /// The type of response format being defined: `json_object`
    JsonObject,
    /// The type of response format being defined: `json_schema`
    JsonSchema {
        json_schema: ResponseFormatJsonSchema,
    },
}
```

might need to be...

```rust
pub enum ResponseFormat {
    /// The type of response format being defined: `text`
    Text,
    /// The type of response format being defined: `json_object`
    JsonObject,
    /// The type of response format being defined: `json_schema`
    JsonSchema(ResponseFormatJsonSchema),
}
```

So that `name`, etc. is at the "top-level" of `text.format`.

samvrlewis:
Thanks for testing it @twitchax and good catch, I think you're right. I updated it to use a `ResponseFormat` just for this API rather than reusing the chat one. The example uses a jsonschema response format now and works for me.

fwiw, I mostly hand generated most of this as I couldn't find any generators that produced nice output. So with how complex the API is, I do worry there might be other subtle cases like this. I suppose if there are more issues like this they can be fixed as encountered though.

twitchax:
@samvrlewis, your changes fixed this issue with parsing the `RepsonseFormat`, and it works for me.  Not that my review matters, but LGTM. 👍

twitchax:
Actually, getting another issue, which just seems like a serialization problem.  When I switch to `o3-mini`, I get...

```
ERROR Error while handling: failed to deserialize api response: missing field `status` at line 18 column 4
```

samvrlewis:
Huh, yeah, there is no status.

```
  "output": [
    {
      "id": "rs_68305c0068848191ab9fa768452d6b090e0209c29f190939",
      "type": "reasoning",
      "summary": []
    },
    {
      "id": "msg_68305c00cac481918b57e1b255671ace0e0209c29f190939",
      "type": "message",
      "status": "completed",
      "content": [
        {
          "type": "output_text",
          "annotations": [],
          "text": "{\"title\": \"Historic Climate Accord Reached: Nations Unite to Combat Global Warming\", \"website\": \"https://www.globalnews.com\"}"
        }
      ],
      "role": "assistant"
    }
  ],
```

Unless I'm misunderstanding them, the docs seem to say that it should have one though..

image

Maybe "populated when items are returned via API" means "not in response to a request"? 🤷‍♂️ 

In any case, I pushed an update to make the field optional so it doesn't make deser. Thanks for the ongoing testing @twitchax!

twitchax:
It looks like `ToolDefinition` may also need an `mcp` option, or, at least, that's what it looks like in the MCP docs.

https://platform.openai.com/docs/guides/tools-remote-mcp

samvrlewis:
> It looks like `ToolDefinition` may also need an `mcp` option, or, at least, that's what it looks like in the MCP docs.
> 
> https://platform.openai.com/docs/guides/tools-remote-mcp

Ah yeah looks like OpenAI added a bunch new tools last week: https://platform.openai.com/docs/changelog !

Have updated the PR to include them, and updated the example to use the same MCP example from the above link.

twitchax:
LGTM after that change.  Had to bump a few types, but all my tests still pass, so... 👍?

Lol.

Haven't tried the MCP stuff, but I bet it works.  May try it in a few weeks.  If you're interested, I'm currently using this for https://github.com/twitchax/triage-bot, which is maybe 70% there?  Mostly works, but needs some tweaking.

twitchax:
@samvrlewis, do you know the best way to respond with type `function_call_output`.

https://platform.openai.com/docs/guides/function-calling?api-mode=responses

Maybe a new enum option needs to be added to `Input`?  It also looks like you're supposed to append the API's tool call message itself, and I'm not sure that's possible either?

Maybe, due to the explosion of options, `Input` just needs a `Custom` state that just lets you shove a raw `serde_json::Value` in there?  Just thinking out loud.

samvrlewis:
> Maybe, due to the explosion of options, Input just needs a Custom state that just lets you shove a raw serde_json::Value in there? Just thinking out loud.

Ooh, yeah, there are a lot of input items that aren't there right now. 😬  Would be nice to have these all strictly typed but for now I've done as you suggested and added a `Custom` variant.

Added another example that uses the `get_weather` example, seems to work!

Thanks for the prompts on MCP, btw! Definitely something to explore a bit further when I have some time. I do worry about how well the model would work with a codebase of significant complexity though, if it's needing to navigate around file by file. I haven't had much success using coding tools like cursor against big repos, I usually need to give them a tighter context to get any good output. Though to be fair I'm usually focused on generating code, maybe just reading/understanding code would be easier?

twitchax:
+1, mileage varies a ton for the "agent mode" stuff like Cursor and Copilot Agent.  I have the same experience with them as you.  Limiting to tests or small refactors appears to work well.  Everything else tends to fold, and produce bad architecture.

For my purposes, MCP is going to be a big deal, so I've been poking at it a little more than others might.

Thanks for entertaining my random observations while I poke!  Happy to help out at some point, but not certain how you and other maintainers feel about stop-gaps like `Custom` enum placeholders.

64bit (approved):
Thank you so much @samvrlewis and @twitchax for your contributions! 

Thank you for doing all the heavy lifting by hand typing types and testing. Appreciate all the hard work! 

Your design choice to nest types inside `types::responses` is a good one!

Issue #264, opened 2024-09-02
Use actively maintained `backon` to replace `backoff`


Xuanwo:
Hi, thanks for building `async-openai`. I'm starting this issue to discuss whether it's a good idea to replace `backoff` with the actively maintained [`backon`](https://github.com/Xuanwo/backon). I noticed that this crate exposes `backoff` at the public API level. Therefore, I want to discuss it first before beginning any work.

---

BackON has now [reached version 1.0](https://xuanwo.io/2024/08-backon-reaches-v1/) and has only one dependency. It can be used in both WebAssembly and `no_std` environments.

I'm willing to help perform the migration if you think it's cool.

64bit:
Hey, thank you for the offer to add support for `backon`, looks really nice project. 

At the moment `backoff` gets the job done.  However, it would be worth revisiting this in the future for WebAssembly and `no_std` environments. We can have this issue parked for now.

mccormickt:
Just wanted to bump this issue now that there is an official advisory regarding `backoff` being unmaintained https://rustsec.org/advisories/RUSTSEC-2025-0012.

Xuanwo:
I no longer have interest in this issue, closing.

PR #375, opened 2025-05-29
replace backoff with backon

spencerbart:
backoff is unmaintained https://rustsec.org/advisories/RUSTSEC-2025-0012 and it is recommended to use backon #264.

This is a breaking change.


TStansel (approved):

64bit:
Thanks for this PR! 

I have been thinking retries and other capabilities and Tower fits really well - and so we dont need external backoff/backon libraries as dependency anymore. 

The v0.37.0 now supports Tower and Tower comes with its own exponential backoff implementation that can be reused. 

Hence closing this.

PR #379, opened 2025-06-02
fix web search options; skip serializing if none

spencerbart:


64bit (approved):
Thank you for the fix!

PR #385, opened 2025-06-04
add flex service tier to chat completions

spencerbart:
https://platform.openai.com/docs/guides/flex-processing?api-mode=chat
https://platform.openai.com/docs/api-reference/chat/create#chat-create-service_tier


64bit (approved):
Thank you for PR!

PR #388, opened 2025-06-10
Add missing voice Ballad to enum

jregistr:
Hey folks. I was parsing a config file based on the voice options here: https://platform.openai.com/docs/guides/text-to-speech#voice-options
and I got an error that "ballad" is not an expected option. I added it to the enum. 

I didn't see any test to update. Let me know if there's anything missing.


openapi.yaml
```diff
@@ -11439,7 +11439,6 @@ components:
                 voice:
                     description: The voice to use when generating the audio. Supported voices are `alloy`, `echo`, `fable`, `onyx`, `nova`, and `shimmer`. Previews of the voices are available in the [Text to speech guide](/docs/guides/text-to-speech/voice-options).
                     type: string
```
64bit, line 11441:
```suggestion
                    type: string
                    enum: ["alloy", "echo", "fable", "onyx", "nova", "shimmer"]
```

PR #393, opened 2025-06-27
task: Add serialize impl for ApiError

tomharmon:
Hi! thanks for the great ilbrary! I would love to see if you'd be open to adding a `Serialize` impl to the `ApiError` type so it can round trip. 

I also bumped the Cargo.toml version so a new crate version could be published, if you are able to do that after merging that would be awesome. 

The diff in `async-openai/src/lib.rs` was from running `cargo fmt`

### DIFF
--- a/async-openai/Cargo.toml
+++ b/async-openai/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "async-openai"
+version = "0.28.3"
 authors = ["Himanshu Neema"]
 categories = ["api-bindings", "web-programming", "asynchronous"]
 keywords = ["openai", "async", "openapi", "ai"]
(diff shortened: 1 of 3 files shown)


async-openai/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "async-openai"
-version = "0.28.3"
+version = "0.28.4"
```
64bit, line 3:
```suggestion
```

async-openai/Cargo.toml
```diff
@@ -1,6 +1,5 @@
 [package]
 name = "async-openai"
```
64bit, line 2:
```suggestion
name = "async-openai"
version = "0.28.3"
```

64bit (approved):
Glad to hear you found it useful! and thank you for the PR! 

Version bump is not required in PR - because that's taken care of by `cargo release` - also there are changes in `main` that calls for a minor version bump instead of patch version bump.

Issue #314, opened 2025-01-11
Spec Sync [Always Open]


64bit:
There are many new changes in spec.  This issue will be open for any future PR to address it on rolling basis. 

Please consider sending PR which syncs spec. 

Each PR scope could be just single API endpoint or single type - to make it easily reviewable and subsequently merged.

PR #402, opened 2025-07-08
feat: update image generation API to match latest OpenAI specs

Madoshakalaka:
## Summary
- Added ImageModeration enum with 'auto' (default) and 'low' values for content-moderation control
- Added moderation parameter to CreateImageRequest for controlling content filtering on gpt-image-1
- Extended ImageQuality enum to support 'high', 'medium', and 'low' values for gpt-image-1 model


Madoshakalaka:
this is not AI slop btw

previously talked about here
https://github.com/64bit/async-openai/issues/314#issuecomment-3042222327

async-openai/src/types/image.rs
```diff
@@ -57,6 +57,9 @@ pub enum ImageQuality {
     #[default]
     Standard,
     HD,
+    High,
+    Medium,
+    Low,
```
64bit, line 62:
```suggestion
    Low,
    Auto,
```

PR #405, opened 2025-07-19
Add streaming support for Responses API.

kazzix14:
Added a streaming support for Responses API.

PR #406, opened 2025-07-24
End stream when Error::StreamEnded appears

kazzix14:
In the Responses API, `event_source.next().await` returns `Err(Error::StreamEnded)` when the stream ends. This PR makes us stop the stream as soon as that error appears. Without this, stream never ends without `break` on userside.

related to https://github.com/64bit/async-openai/pull/405.


64bit:
There's work on streaming error in a different direction https://github.com/64bit/async-openai/pull/413

PR #412, opened 2025-08-11
Add skip_serializing_if to more option types

howardjohn:
These are already used throughout the project, just inconsistently


64bit:
It was only on types serialized, hence you won't find it on response types

howardjohn:
Serializing response types is something I am doing for auditing, testing, etc

PR #413, opened 2025-08-12
Unify StreamError and OpenAIError

tinco:
The current StreamError implementation wraps whatever happens in a string, throwing away important underlying information. Specifically this prevents us from catching ContextLengthExceeded errors when streaming is enabled.

This PR rewrites the error handling so that the streaming and non-streaming implementations share the same error parsing code and both return OpenAIError's of which StreamError is now a stream specific subset.

Unfortunately this is a breaking change, any users that explicitly match on StreamError will have to be modified to accommodate the extra information that's now returned in the error.

### DIFF
--- a/async-openai/src/error.rs
+++ b/async-openai/src/error.rs
@@ -4,7 +4,6 @@ use std::string::FromUtf8Error;
 use reqwest::{header::HeaderValue, Response};
 use serde::{Deserialize, Serialize};
 
-use reqwest_eventsource::Error as EventSourceError;
 
 #[derive(Debug, thiserror::Error)]
 pub enum OpenAIError {
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -346,19 +346,17 @@ impl<C: Config> Client<C> {
                                     err: OpenAIError::ApiError(api_error),
                                     retry_after: None,
                                 })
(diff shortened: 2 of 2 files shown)


async-openai/src/error.rs
```diff
@@ -20,13 +24,35 @@ pub enum OpenAIError {
     FileReadError(String),
     /// Error on SSE streaming
     #[error("stream failed: {0}")]
-    StreamError(String),
+    StreamError(StreamError),
```
timonv, line 27: Nice! Since you always map the inner eventsource error anyway, It would simplify the code quite a bit I think to wrap StreamError(EventSourceError) instead with #[from].

If I'm not mistaken, that would remove `map_stream_error` instead as we can use Into, remove the intermediate error, and avoid reading out the bytes of the response early.

tinco: We can't because we need to determine what `OpenAIError` is wrapped inside the `EventSourceError::InvalidStatusCode` and as a bonus we also map the `EventSourceError::Transport` to `OpenAIError::Reqwest`.

64bit (changes requested):
Don't think this library should unwrap and match another library's error structure - please consider achieveing same outcome with `OpenAIError::StreamError(EventSourceError)`

PR #414, opened 2025-08-12
Add ReasoningEffort::Minimal

MakotoE:
The Minimal enum value for ReasoningEffort is added. The newest gpt-5 model supports this value for reasoning_effort. [Docs](https://platform.openai.com/docs/api-reference/chat/create#chat_create-reasoning_effort)


64bit:
Thank you for the PR - looks duplicate of https://github.com/64bit/async-openai/pull/411

PR #415, opened 2025-08-14
feat: Respect backoff policy for streaming

timonv:
Went into the rabbit hole of streaming backoffs. Turns out eventsource has something already, of which I'm not 100% certain how well it works, but since streaming is a bit different, we want to generally retry all the time. This PR sets the event source retry policy to the (similar) values of backoff. There's one slight difference in behaviour, in that on transport errors, we only retry server errors or 429s, like with regular requests.

We can't inspect the response body at this point, so we can't differentiate between rate limits and quota errors (thanks openai).

Also, during streaming, token errors are a rough 400. I think maybe. Sometimes.

Practically, I'm trying to stream a lot of agents in parallel, hoping with this I can mitigate it somewhat.

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -3,7 +3,7 @@ use std::pin::Pin;
 use bytes::Bytes;
 use futures::{stream::StreamExt, Stream};
 use reqwest::multipart::Form;
-use reqwest_eventsource::{Event, EventSource, RequestBuilderExt};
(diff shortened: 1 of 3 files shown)


timonv:
Actually, #413 would be a nice extension to this

64bit:
hey after recent refactor the default retry behaviour actually addresses 429, 5xx, and connection errors already.

Also we had to remove reqwest_eventsource dependency to upgrade to reqwest 0.13. 

If you think above does not cover your original use case - please open an issue!

PR #418, opened 2025-08-19
Baseten.co: Add reasoning field for reasoning parser + grammar mode

michaelfeil:
Hey,

we @basetenlabs / @nvidia are using your crate to run validations and output return. 

Two things are missing in the Rust crate over e.g. the Python types. 

- https://github.com/openai/openai-python/blob/4ada66f8f86473f342aa032ed021b62180422dc1/src/openai/types/shared/response_format_text_grammar.py#L10 `grammar` field. 
- adds the reasoning field for the reasoning content https://openrouter.ai/docs/use-cases/reasoning-tokens


64bit:
Thanks for the PR, however please see contribution guidelines for the scope of crate. 

You can achieve this with combination of `byot` feature and Serde https://serde.rs/attr-flatten.html#struct-flattening

64bit:
Its in the readme: https://github.com/64bit/async-openai?tab=readme-ov-file#contributing

PR #419, opened 2025-08-20
fix: Add Scale and Priority to the `ServiceTier` enum for the Responses API

Advayp:
## Description
- This PR adds the Scale and Priority variants to the ServiceTier enum, similar to #416 
- However, #416 did so only for the Chat Completions API. This PR adds these variants for the Responses API.


64bit (approved):
Thank you for the fix!

PR #426, opened 2025-08-26
fix: Update `OutputItem` to align with OpenAI's Specification

Advayp:
## Description
- According to OpenAI's [specification](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output), `OutputItem` should be a tagged union. This PR updates that struct to align with OpenAI's specification
- A nice benefit of this is that streaming with tool calls is now supported

### DIFF
--- a/async-openai/src/types/responses.rs
+++ b/async-openai/src/types/responses.rs
@@ -2104,20 +2104,34 @@ pub struct ResponseMetadata {
 
 /// Output item
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


64bit (changes requested):
Looks like this PR is updating incorrect type, the `Response` object output field type is `OutputContent` 

https://github.com/64bit/async-openai/blob/main/async-openai/src/types/responses.rs#L1371

64bit (approved):
Thank you for clarification and PR! it makes sense to update OutputItem type with spec

PR #429, opened 2025-09-02
fix: CompoundFilter should use CompoundType instead

the-spice-must-flow:
The response filters had a typo/mistake where the wrong comparison type is used in a compound filter.

We want options like AND or OR not Equals, NEquals etc.

https://platform.openai.com/docs/api-reference/responses/create

Screenshot 2025-09-02 at 19 12 36


64bit (approved):
Thank you for the fix!

PR #439, opened 2025-10-06
responses: add verbosity option to TextConfig

joune:
Minimalist patch to add support for verbosity config


64bit (approved):
Thank you for updating spec!

PR #442, opened 2025-10-13
feat(spec): Add `name` field to `ResponseFunctionCallArgumentsDone`

surafelfikru:
Added a name field to the ResponseFunctionCallArgumentsDone struct. Its crucial  to identify the tool OpenAI instructed to execute, allowing users to filter and select the appropriate tool for execution. This change also syncs to OpenAI specs.


64bit (approved):
Thank you!

PR #444, opened 2025-10-14
Add support for expires_after field in CreateFileRequest and expires_at field in OpenAIFile

MakotoE:
- The `expires_after: Option<FileExpiresAfter>` field is added to `CreateFileRequest`. This new field is added to the `post https://api.openai.com/v1/files` request when defined.
  - https://platform.openai.com/docs/api-reference/files
  - I tried to follow the patterns of other fields by adding the `FileExpiresAfterAnchor` enum, even though it only contains one option, `CreatedAt`.
- The `expires_at: Option<u32>` field is added to OpenAIFile. 
  - https://platform.openai.com/docs/api-reference/files/object
  - I found that `expires_at` can be `null` if you didn't specify `expires_after` in the request, even though the OpenAI documentation does not describe it as being optional. That is why I made the type `Option<u32>`, not `u32`.

### DIFF
--- a/async-openai/src/file.rs
+++ b/async-openai/src/file.rs
@@ -70,7 +70,7 @@ impl<'c, C: Config> Files<'c, C> {
 #[cfg(test)]
 mod tests {
     use crate::{
-        types::{CreateFileRequestArgs, FilePurpose},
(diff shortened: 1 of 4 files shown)


64bit (approved):
Thank you for spec update!

PR #450, opened 2025-10-21
style: add explicit lifetime parameter to methods

moldhouse:
with changes to rust lifetime elision rules, the compiler generates a warning for methods that return a type that has a lifetime parameter tied to &self but are not explicit about it


64bit (approved):
Thank you for the PR!

PR #451, opened 2025-10-21
feat!: remove api_key and api_base from config trait

moldhouse:
when creating a custom config, it is not clear why these methods would need to be implemented


64bit:
It was introduced to work with Azure, and useful for other openai compatible apis

Issue #448, opened 2025-10-20
Bring Your Own Error Type


moldhouse:
When using the BYOT features with third-party providers, the error format of these APIs often does not match the expectation of this library. For example, this library deserializes errors into this format:

```rs
pub struct ApiError {
    pub message: String,
    #[serde(rename = "type")]
    pub r#type: Option<String>,
    pub param: Option<String>,
    #[serde(deserialize_with = "deserialize_code")]
    pub code: Option<String>,
}
```

However, OpenRouter provides errors in the following format:

```sh
{
    "error": {
        "message": "User not found.",
        "code": 401
    }
}
```

with the small difference that code is an integer, leading to a deserialization error when using this library.

Before implementing a possible "bring-your-own-error-type" feature, I want to check if this would be within the scope of this library, and about how to architect it. One option might be that users bring their own error type for deserialization that must implement a trait to convert itself into an `OpenAIError`.

64bit:
In `v0.37.0` now supports Tower middlewares and custom error types in `OpenAIError::Boxed( ... )` variant. So one can leverage it for custom error types and then downcast it if type is known. 

Hence closing this

PR #452, opened 2025-10-22
String errors

moldhouse:
References #448 

In the issue, the idea of exposing a `serde_json::Value` was discussed. In some internal server error cases, even `OpenAI` does return error messages that are not json. This made me wonder if representing the error as `String` might give users more flexibility. Would be eager to get some feedback if this matches what you had in mind @64bit 

I'm uncertain if the second commit (offering a helper method to parse the raw string into a type) provides value / goes beyond the scope of this library, happy to remove it.


64bit:
Thanks again for this - this is now supported for any custom error types via https://docs.rs/async-openai/latest/async_openai/error/enum.OpenAIError.html#variant.Boxed

Hence closing.

PR #459, opened 2025-11-05
Use Retry-After if present in header

kczimm:
When using the `backoff::future::retry`, we can attempt to parse and use the `Retry-After` header value.


64bit:
Alternatively introducing `RetryOptions` which contains backoff and a header name for "retry-after" could work. If you'd like to tackle it!

64bit:
Thanks for this, actually crate now honors retry-after header when present by default, hence closing this PR.

PR #500, opened 2025-12-12
fix: improve example output format

jacoesad:
- azure-openai-service
- chat
- ollama-chat


64bit:
Thanks but crate is already `cargo fmt`

PR #528, opened 2026-03-17
Add tool_search support for gpt5

monadoid:
gpt5 models have a new tool_search functionality - this adds support for it.

Updates `openapi.documented.yml` to the latest official documented OpenAI spec from:
https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml

The OpenAPI/API changes are additive, but the typed Rust Responses surface now includes new public struct fields and enum variants which would be a breaking change for users using the struct literals (not builders) or doing exhaustive matching.

Adds typed Responses API support for `tool_search`, including:
- `namespace`
- deferred namespace-contained function tools
    - `tool_search_call`
    - `tool_search_output`
    - `namespace` on function call items

Added tests + "client mode" and "hosted mode" examples and confirmed working.


64bit:
Thanks for the PR, however syncing spec requires to sync whole spec, because thats the source of truth on what code reflects. 

For reference, this are all the changes required to sync spec: https://github.com/64bit/async-openai/pull/529 


You're welcome to cleanup this PR and keep a single example for tool search.

PR #518, opened 2026-01-26
chore: update to reqwest 0.13

ttys3:
since this crate depends on reqwest-eventsource

we still need to wait reqwest-eventsource upgrade to reqwest 0.13

see https://github.com/jpopesculian/reqwest-eventsource/pull/35

PR #537, opened 2026-04-23
chore: upgrade reqwest to 0.13 and remove reqwest-eventsource

richarddd:
Bumps reqwest from 0.12 to 0.13. `reqwest-eventsource` is abandoned and has no 0.13 release, so rather than pull in the `aha-reqwest-eventsource` fork (suggested in #518), this drops the dependency entirely.

The multipart streaming path was already consuming SSE via `reqwest::Response::bytes_stream()` + the `eventsource-stream` crate; the other three streaming call sites (`post_stream`, `post_stream_mapped_raw_events`, `get_stream`) now use the same pattern. `eventsource-stream` was already a direct dependency, so this is one fewer crate in the tree.

Also drops the `rustls-webpki-roots` feature: reqwest 0.13 unified TLS roots behind the `rustls` feature (via `rustls-platform-verifier`), and the webpki variant is no longer a distinct upstream option. Users who need explicit webpki roots can configure them through reqwest's `ClientBuilder`.

Net: -80 LOC, no new deps.

### DIFF
--- a/async-openai/Cargo.toml
+++ b/async-openai/Cargo.toml
@@ -14,10 +14,8 @@ repository = "https://github.com/64bit/async-openai"
 
 [features]
 default = ["rustls"]
-# Enable rustls for TLS support
-rustls = ["dep:reqwest", "reqwest/rustls-tls-native-roots"]
-# Enable rustls and webpki-roots
-rustls-webpki-roots = ["dep:reqwest", "reqwest/rustls-tls-webpki-roots"]
+# Enable rustls for TLS support (uses rustls-platform-verifier for roots in reqwest 0.13)
+rustls = ["dep:reqwest", "reqwest/rustls"]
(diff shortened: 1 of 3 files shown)


64bit (approved):
Nice work! thank you for making this happen. 

~This also would cleanup code related to error handling like wasm/non-wasm - I can do that as followup.~

~With this I think streaming should work for wasm too~ (edit: this is related to tokio and stuff my bad)

richarddd:
> Nice work! thank you for making this happen.
> 
> ~This also would cleanup code related to error handling like wasm/non-wasm - I can do that as followup.~
> 
> ~With this I think streaming should work for wasm too~ (edit: this is related to tokio and stuff my bad)

Thanks for fast review and for keeping this project alive!!! Hats off 🫡

PR #538, opened 2026-04-26
`From<OutputItem> for Item` (and `InputItem`) for multi-turn responses

### LINKED ISSUE #492: Multi-turn Responses API with Reasoning
Typically, if I was using the responses API in python, I would add all output items from the responses API into some ongoing history list.

```
response = client.responses.create(
    model="gpt-5",
    tools=tools,
    input=input_list,
)

# Save function call outputs for subsequent requests
input_list += response.output
```

This ensures critical fields like reasoning stay in model input and allow caching to work.

However, none of the examples of the responses API really show how to do this, and I'm hitting a lot of type compability issues between InputItem and OutputItem, and its causing the model history to drop important fields like reasoning. 

The existing examples of the responses API only show manually constructing function call items and appending to the history list, but it'd be much nicer if there was a way to do something like `history.extend(response.output.into())` to convert vec[OutputItem] into vec[InputItem]

Madoshakalaka:
Closes #492

My code worked with 5.4 and started breaking with 5.5 and ~~I~~ Opus 4.7 did some investigation, I pulled the official python and node SDKs and had a look.

## Problem

When a reasoning model (gpt-5 etc.) is used through the Responses API with tool calls, the model's `response.output` contains `Reasoning` items alongside the `FunctionCall` items they preceded. The API requires those reasoning items to be echoed back into the next request's `input` together with the function calls. Otherwise the server rejects the second turn:

```
Item 'fc_...' of type 'function_call' was provided without its required 'reasoning' item: 'rs_...'
```

This is a recent server-side enforcement change (the same code worked previously, and the existing `responses-function-call` example silently has the same bug for a reasoning model — it only echoes back the `FunctionCall`).

In the official Python and Node SDKs the canonical pattern is one line:

```python
input_list += response.output  # python
```

```ts
inputItems.push(...response.output);  // node
```

both work because their input/output item types are the same. In Rust the OpenAPI-derived schemas split into `Item` (input side) and `OutputItem` (output side), and several variants have `*Resource` (output, required `id`/`status`) vs `*ItemParam` (input, optional `id`/`status`) flavors. So callers need a per-variant conversion to do what's a one-liner in the official SDKs — and most users don't realize they need to do it until the API rejects their request.

### DIFF
--- a/async-openai/src/types/responses/impls.rs
+++ b/async-openai/src/types/responses/impls.rs
@@ -1,20 +1,34 @@
 use crate::types::mcp::MCPTool;
 use crate::types::responses::{
-    ApplyPatchToolCallItemParam, ApplyPatchToolCallOutputItemParam, CodeInterpreterContainerAuto,
-    CodeInterpreterTool, CodeInterpreterToolCall, CodeInterpreterToolContainer,
-    ComputerCallOutputItemParam, ComputerTool, ComputerToolCall, ComputerUsePreviewTool,
-    ConversationParam, CustomToolCall, CustomToolCallOutput, CustomToolParam, EasyInputContent,
-    EasyInputMessage, FileSearchTool, FileSearchToolCall, FunctionCallOutput,
-    FunctionCallOutputItemParam, FunctionShellCallItemParam, FunctionShellCallOutputItemParam,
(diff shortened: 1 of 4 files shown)


64bit (approved):
Wow that was non trivial - thank you for adding proper multi-turn support and example!

PR #543, opened 2026-05-11
fix: check [DONE] SSE marker before deserialization in stream_mapped_raw_events

### LINKED ISSUE #542: bug: [DONE] SSE marker fails deserialization in stream() after v0.36.0 refactor
## Description

In v0.36.0 (PR #537), the streaming pipeline was refactored to remove the `reqwest-eventsource` dependency. As part of this change, the `stream()` function was rewritten to delegate to `stream_mapped_raw_events`.

The old `stream()` function (v0.36.1) checked for the `[DONE]` SSE end-of-stream marker **before** passing the event data to `serde_json::from_str`:

https://github.com/64bit/async-openai/blob/v0.36.1/src/client.rs#L708-L716
```rust
if event.data == "[DONE]" {
    break;  // ← exits before deserialization
}
let response = serde_json::from_str::<O>(&event.data)...;
```

The new `stream()` function (v0.38.0) delegates to `stream_mapped_raw_events`, which checks `[DONE]` **after** calling the `event_mapper` closure:

https://github.com/64bit/async-openai/blob/v0.38.0/src/client.rs#L814-L821
```rust
let done = event.data == "[DONE]";  // ← flag is stored
let response = event_mapper(event); // ← deserialization happens BEFORE done check
return Some((response, (event_stream, event_mapper, done)));
```

Since the `event_mapper` for `stream()` is `|event| serde_json::from_str::<O>(&event.data)`, the `[DONE]` string gets passed directly to the JSON parser, resulting in:

```
failed deserialization of: [DONE]
error: expected value at line 1 column 2
```

This affects both the WASM and non-WASM implementations of `stream_mapped_raw_events`.

## Expected Behavior

`[DONE]` is a valid SSE protocol-level end-of-stream marker. It should terminate the stream cleanly without being passed to the deserializer.

## Steps to Reproduce

1. Use `async-openai` v0.38.0
2. Call `Chat::create_stream_byot()` with any custom deserialized type
3. The stream will produce an error when the server sends `data: [DONE]`

## Proposed Fix

Move the `[DONE]` check before the `event_mapper` call in both implementations of `stream_mapped_raw_events`, restoring the behavior from v0.36.1.

Happy to submit a PR if this looks correct.

Linuxdazhao:
## Summary

Fixes #542 — `[DONE]` SSE end-of-stream marker being passed to `serde_json::from_str` in `stream()`, causing a deserialization error.

## Root Cause

In v0.36.0 (PR #537), the streaming pipeline was refactored to remove `reqwest-eventsource`. The old `stream()` function checked for `[DONE]` **before** deserialization. The new `stream()` delegates to `stream_mapped_raw_events`, which checks `[DONE]` **after** the mapper closure — and the mapper in `stream()` is `serde_json::from_str`.

## Fix

Move the `[DONE]` check before the `event_mapper` call in both implementations:

- **WASM** (`futures::stream::unfold`): Check `[DONE]` and return `None` to end the stream cleanly. Removed the `finished` state variable since it's no longer needed.
- **Non-WASM** (`tokio::mpsc`): Check `[DONE]` and `break` before calling `event_mapper`.

## Test Plan

- Existing unit tests pass (15 passed, 11 skipped — require API key)
- Change is minimal and restores the pre-refactor behavior

## Before

```rust
let done = event.data == "[DONE]";
// ...
let response = event_mapper(event); // [DONE] passed to JSON parser → error!
return Some((response, (event_stream, event_mapper, done)));
```

## After

```rust
if event.data == "[DONE]" {
    return None; // clean stream termination
}
// ...
let response = event_mapper(event);
return Some((response, (event_stream, event_mapper)));
```

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -791,12 +791,8 @@ where
     let event_stream = Box::pin(eventsource_stream::EventStream::new(byte_stream));
 
     Box::pin(futures::stream::unfold(
-        (event_stream, event_mapper, false),
-        |(mut event_stream, event_mapper, finished)| async move {
-            if finished {
-                return None;
-            }
-
+        (event_stream, event_mapper),
+        |(mut event_stream, event_mapper)| async move {
             loop {
                 let event = match event_stream.next().await {
                     Some(Ok(event)) => event,
@@ -805,20 +801,22 @@ where
                             Err(OpenAIError::StreamError(Box::new(
                                 StreamError::EventStream(error.to_string()),
                             ))),
-                            (event_stream, event_mapper, true),
+                            (event_stream, event_mapper),
                         ));
                     }
                     None => return None,
                 };
 
-                let done = event.data == "[DONE]";
+                if event.data == "[DONE]" {
+                    return None;
+                }
 
                 if event.event == "keepalive" {
                     continue;
                 }
 
                 let response = event_mapper(event);
-                return Some((response, (event_stream, event_mapper, done)));
+                return Some((response, (event_stream, event_mapper)));
             }
         },
     ))
@@ -856,7 +854,9 @@ where
                     break;
                 }
             };
-            let done = event.data == "[DONE]";
+            if event.data == "[DONE]" {
+                break;
+            }
 
             if event.event == "keepalive" {
                 continue;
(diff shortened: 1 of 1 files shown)


64bit:
I was trying to reduce duplicate functions - good catch and thanks for the fix!

PR #546, opened 2026-05-13
fix(responses): default missing InputImageContent.detail during deserialization

### LINKED ISSUE #545: Multimodal Responses payload without `detail` fails to deserialize (InputImageContent.detail missing `#[serde(default)]`)
## Summary

`async_openai::types::responses::InputImageContent.detail` is `ImageDetail` with no `#[serde(default)]`. The doc-comment states *"Defaults to `auto`"*, but in practice any spec-compliant payload that omits the field fails to deserialize. Because `InputItem` is `#[serde(untagged)]`, the resulting `missing field 'detail'` error rolls up to the unhelpful `data did not match any variant of untagged enum InputItem`.

Affects v0.38.1 and current `main` (`async-openai/src/types/responses/response.rs:704-715`).

## Repro

```rust
use async_openai::types::responses::InputItem;
use serde_json::json;

fn main() {
    let payload = json!({
        "role": "user",
        "content": [
            {"type": "input_text", "text": "describe this"},
            {"type": "input_image", "image_url": "https://example.com/cat.png"}
        ]
    });
    let result: Result<InputItem, _> = serde_json::from_value(payload);
    println!("{:?}", result);
}
```

**Actual (v0.38.1):**
```
Err(Error("data did not match any variant of untagged enum InputItem", line: 0, column: 0))
```

**Expected:** `Ok(InputItem::EasyMessage(...))` with the `input_image` part's `detail` defaulted to `ImageDetail::Auto`.

## Why this matters

The OpenAI Responses API spec and every official SDK treat `detail` as optional with default `auto`:

- **openai-python** — [`ResponseInputImageParam`](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_input_image_param.py): `detail: Optional[Literal[\"low\", \"high\", \"auto\"]]`
- **openai-node** — `responses.ts`: `detail?: 'low' | 'high' | 'auto'`

Any spec-compliant client that doesn't explicitly send `detail` on every `input_image` part hits this. We hit it downstream in [ai-dynamo/dynamo#9470](https://github.com/ai-dynamo/dynamo/pull/9470) (had to shadow `EasyInputMessage` + `EasyInputContent` + `InputImageContent` as a workaround); the error was non-obvious to debug because the untagged-enum wrap-up hides the actual missing-field cause.

## Root cause

`InputImageContent.detail` lacks `#[serde(default)]`. `ImageDetail` already derives `Default` with `#[default] Auto`, so the fix is a one-attribute change.

## Proposed fix

```diff
 pub struct InputImageContent {
     /// The detail level of the image to be sent to the model. One of \`high\`, \`low\`, or \`auto\`.
-    /// Defaults to \`auto\`.
+    /// Defaults to \`auto\` when omitted in JSON input.
+    #[serde(default)]
     pub detail: ImageDetail,
```

Mirrors the precedent set by #527 which fixed the same class of bug on `EasyInputMessage.r#type` and `WebSearchApproximateLocation.r#type`.

I have a PR ready with this fix + two focused integration tests appended to `tests/responses_input_item_serde.rs` (one for the `EasyMessage` path, one for the strict `Item::Message(Input::User)` path). All existing tests + new tests pass; `cargo fmt --check` and `cargo clippy -D warnings` clean.

End-to-end reproducer (released v0.38.1 fails with the exact error above; patched build succeeds with `detail` defaulted) available on request.

ajcasagrande:
- default `InputImageContent.detail` to `auto` when missing in JSON input
- add focused response-types integration tests covering both the
  `EasyMessage` and strict `Item::Message(Input::User)` deserialization paths

Closes #545.

`InputImageContent.detail` is documented as defaulting to `auto`, but without
`#[serde(default)]` any spec-compliant payload that omits `detail` (per
openai-python's `ResponseInputImageParam`, which marks `detail`
`Optional[Literal["low","high","auto"]]`) fails to deserialize. Because
`InputItem` is `#[serde(untagged)]`, the single missing-field error rolls up
to `data did not match any variant of untagged enum InputItem`, which is
what spec-compliant clients hit in practice (full traceback in #545).

This mirrors the precedent set by #527 (`EasyInputMessage.r#type`,
`WebSearchApproximateLocation.r#type`) and uses the same test file
(`tests/responses_input_item_serde.rs`).

`ImageDetail` already derives `Default` with `#[default] Auto`, so the
fix is a single `#[serde(default)]` attribute — no `Default` impl needed.

## Verification

- `cargo test -p async-openai --test responses_input_item_serde --features response-types` → 4 passed (2 pre-existing + 2 new)
- `cargo fmt --check` → clean
- `cargo clippy -p async-openai --features response-types --tests -- -D warnings` → clean
- End-to-end repro: against released v0.38.1 from crates.io, the payload below errors with `data did not match any variant of untagged enum InputItem`; against this branch the same payload deserializes successfully with `detail` defaulted to `ImageDetail::Auto`.

```json
{
  "role": "user",
  "content": [
    {"type": "input_text", "text": "describe this"},
    {"type": "input_image", "image_url": "https://example.com/cat.png"}
  ]
}
```

Refs:
- openapi.documented.yml: `ResponseInputImage` (type required; `image_url` / `file_id` / `detail` all optional)
- openai-python: [`src/openai/types/responses/response_input_image_param.py`](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_input_image_param.py) — `detail: Optional[Literal["low", "high", "auto"]]`
- openai-node: `src/resources/responses/responses.ts` — `detail?: 'low' | 'high' | 'auto'`
- Prior fix in same spirit: #527

### DIFF
--- a/async-openai/src/types/responses/response.rs
+++ b/async-openai/src/types/responses/response.rs
@@ -703,7 +703,8 @@ pub struct InputTextContent {
 #[builder(build_fn(error = "OpenAIError"))]
 pub struct InputImageContent {
     /// The detail level of the image to be sent to the model. One of `high`, `low`, or `auto`.
-    /// Defaults to `auto`.
+    /// Defaults to `auto` when omitted in JSON input.
+    #[serde(default)]
     pub detail: ImageDetail,
     /// The ID of the file to be sent to the model.
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/async-openai/tests/responses_input_item_serde.rs
+++ b/async-openai/tests/responses_input_item_serde.rs
@@ -1,8 +1,8 @@
 #![cfg(feature = "response-types")]
 
 use async_openai::types::responses::{
-    EasyInputContent, InputItem, MessageType, Role, WebSearchApproximateLocation,
-    WebSearchApproximateLocationType,
+    EasyInputContent, ImageDetail, InputContent, InputItem, InputRole, Item, MessageItem,
+    MessageType, Role, WebSearchApproximateLocation, WebSearchApproximateLocationType,
 };
 use serde_json::json;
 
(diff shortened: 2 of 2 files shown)


64bit (approved):
Thanks for details and tests, appreciate it all! 

PS: I still see `detail` to be [`Required`](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_input_image_param.py)

PR #550, opened 2026-05-20
fix: Make `action` optional in `response.output_item.added`

### LINKED ISSUE #548: Responses stream fails to deserialize in-progress web_search_call without `action`
OpenAI will sometimes emit `response.output_item.added` events for web searches before the `action` details are populated. This in turn causes an error from `async-openai` as it expects the action field to exist. The payload we've seen come back from OpenAI looks like:

```json
{
  "type": "response.output_item.added",
  "sequence_number": 2,
  "output_index": 0,
  "item": {
    "id": "ws_123",
    "type": "web_search_call",
    "status": "in_progress"
  }
}
```

The lifecycle shape of these events seems to be a bit looser than the final streaming shape. The openai docs describe `web_search_call` with idd, action, status, and type, but this doesn't seem accurate for in-progress events. OpenAIs own SDKs seem to have matching reports for this as well: 
- https://github.com/openai/openai-java/issues/633
  - https://github.com/openai/openai-java/issues/600
- https://github.com/openai/openai-python/issues/2448 

This behavior tested on `async-openai: 0.34.0`. 

A mini reproduction:

```rs
use async_openai::types::responses::ResponseStreamEvent;
use serde_json::json;

fn main() {
    let payload = json!({
        "type": "response.output_item.added",
        "sequence_number": 2,
        "output_index": 0,
        "item": {
            "id": "ws_123",
            "type": "web_search_call",
            "status": "in_progress"
        }
    });

    let err = serde_json::from_value::<ResponseStreamEvent>(payload).unwrap_err();
    println!("{err}");
}
```

with

```
[dependencies]
async-openai = { version = "0.34.0", features = ["responses"] }
serde_json = "1"
```

will render ``missing field `action`` `

rijkvanzanten:
This makes the `action` field of the `WebSearchToolCall` optional to workaround an OpenAI bug.

Fixes https://github.com/64bit/async-openai/issues/548

### DIFF
--- a/async-openai/src/types/responses/response.rs
+++ b/async-openai/src/types/responses/response.rs
@@ -2034,7 +2034,12 @@ pub enum WebSearchToolCallAction {
 pub struct WebSearchToolCall {
     /// An object describing the specific action taken in this web search call. Includes
     /// details on how the model used the web (search, open_page, find, find_in_page).
-    pub action: WebSearchToolCallAction,
+    ///
+    /// This is optional because `response.output_item.added` events can include
+    /// in-progress web search calls before OpenAI has populated the action.
+    /// See <https://github.com/64bit/async-openai/issues/548>.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub action: Option<WebSearchToolCallAction>,
     /// The unique ID of the web search tool call.
     pub id: String,
     /// The status of the web search tool call.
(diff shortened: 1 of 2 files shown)


rijkvanzanten:
@64bit Here ya go! Happy to make alterations if need be. Thanks for the quick response on #548 :) Glad to be able to drop my hacky userland workaround once this is live~

64bit (approved):
Looks good, thank you!

PR #552, opened 2026-05-20
Return http status code for ApiError

### LINKED ISSUE #549: Preserve upstream HTTP status code in `OpenAIError` for non-2xx responses
## Summary

  When an upstream HTTP response is not 2xx, `read_response` ([`src/client.rs#L726-L749`](https://github.com/64bit/async-openai/blob/main/src/client.rs)) extracts the `ApiError` body but drops the original `reqwest::StatusCode`. Callers receive `OpenAIError::ApiError(ApiError { message, r#type, param, code })` with no way to know whether the upstream returned 400, 403, 422, 500, etc. — the only signal is whatever happens to be in the body, which varies wildly across OpenAI-compatible servers (sglang, vLLM, Modal-hosted endpoints, in-house gateways, etc.).

  ## Reproducer

  Any OpenAI-compatible upstream that returns a 4xx whose error body does **not** populate `type` or `code` in the way OpenAI itself does. Example body we see from one of our inference servers on a validation failure:

  ```json
  {
    "error": {
      "message": "Generation failed: temperature must be non-negative, got -1.0.",
      "type": "internal_error",
      "param": null,
      "code": null
    }
  }
  ```

  Whether the server returned this with HTTP 400 or 500, the caller gets the same `ApiError`. We can't route the response back to the client with the correct status because the status is gone.

  ## Why this matters

  We use async-openai inside a gateway service that proxies requests to multiple upstream model providers. When inference returns a 400, we want to surface a 400 to our client (it's their bad request); when inference returns a 500, we want to surface a 500. Today we have to guess by inspecting `ApiError::r#type` against a hand-maintained allowlist of strings (`invalid_request_error` → 400, `authentication_error` → 401, …) and treat anything unrecognized as 500. This is brittle and loses information for any upstream that doesn't follow OpenAI's exact `type` taxonomy.

  Specifically:
  - For 5xx (`read_response` lines 731-741), the body might not even be JSON, so `ApiError` is synthesized with `type: None, code: None` — fully stripping the status.
  - For non-2xx-non-5xx (`read_response` lines 744-748), the body is parsed but the `response.status()` value at line 727 is thrown away.

  ## Proposed change

  When the error originates from a real HTTP exchange, propagate the upstream `StatusCode` through to the caller in some form, so consumers don't have to reconstruct it by string-matching on `type`/`code`. I don't want to prescribe a specific shape (field on `ApiError`, new `OpenAIError` variant, accessor method, etc.) — happy to defer to whatever fits the crate's existing conventions and back-compat policy.

  Happy to send a PR once there's agreement on the shape.

max-zheng:
Preserves the HTTP `StatusCode` from non-2xx responses by changing `OpenAIError::ApiError` to wrap a new `ApiErrorResponse { status_code, api_error }` struct. This lets callers distinguish 400 / 404 / 5xx without pattern-matching on `ApiError.r#type` strings — the original motivation in #549, e.g. for gateways that proxy to non-OpenAI-compatible upstreams.

  Implements option (2) from the issue discussion (the maintainer's preferred shape — wrapping in a struct rather than a struct-variant, so a single value can be passed/extended later).

  This is a **breaking change** for callers that pattern-match on `OpenAIError::ApiError(_)`: the inner type is now `ApiErrorResponse` rather than `ApiError`, accessed via `.api_error`. Per the issue thread, a dedicated minor release was suggested.

  Closes #549.

  ## Changes
  - `error.rs`: add `ApiErrorResponse` (status + api_error, with `Display`/`Error` impls); change variant to `ApiError(ApiErrorResponse)`. The existing `ApiError` struct and its JSON-body `Display` are untouched.
  - `client.rs` `read_response`: wrap both 5xx and non-2xx construction sites with the captured `status`.
  - `middleware/retry/openai.rs`: capture `status_code` before consuming the response body in the 429 path; wrap both the transient and `insufficient_quota` error returns.

### DIFF
--- a/async-openai/src/error.rs
+++ b/async-openai/src/error.rs
@@ -114,7 +114,9 @@ impl std::error::Error for ApiError {}
 #[cfg(feature = "_api")]
 #[derive(Debug, Clone)]
 pub struct ApiErrorResponse {
+    /// HTTP status code
     pub status_code: reqwest::StatusCode,
+    /// Parsed error from response 
     pub api_error: ApiError,
 }
 
--- a/async-openai/src/error.rs
+++ b/async-openai/src/error.rs
@@ -116,7 +116,7 @@ impl std::error::Error for ApiError {}
 pub struct ApiErrorResponse {
     /// HTTP status code
     pub status_code: reqwest::StatusCode,
-    /// Parsed error from response 
+    /// Parsed error from response
     pub api_error: ApiError,
 }
 
--- a/async-openai/src/error.rs
+++ b/async-openai/src/error.rs
@@ -8,9 +8,10 @@ pub enum OpenAIError {
     /// Underlying error from reqwest library after an API call was made
     #[error("http error: {0}")]
     Reqwest(#[from] reqwest::Error),
-    /// OpenAI returns error object with details of API call failure
+    /// OpenAI returns error object with details of API call failure, along
+    /// with the HTTP status code from the response.
     #[error("{0}")]
-    ApiError(ApiError),
+    ApiError(ApiErrorResponse),
     /// Error when a response cannot be deserialized into a Rust type
     #[error("failed to deserialize api response: error:{0} content:{1}")]
     JSONDeserialize(serde_json::Error, String),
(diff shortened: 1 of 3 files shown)


async-openai/src/error.rs
```diff
@@ -109,6 +110,24 @@ impl std::fmt::Display for ApiError {
 
 impl std::error::Error for ApiError {}
 
+/// `ApiError` paired with the HTTP status code from the response.
+#[cfg(feature = "_api")]
+#[derive(Debug, Clone)]
+pub struct ApiErrorResponse {
+    pub status_code: reqwest::StatusCode,
+    pub api_error: ApiError,
```
64bit, line 120:
```suggestion
    /// HTTP status code
    pub status_code: reqwest::StatusCode,
    /// Parsed error from response 
    pub api_error: ApiError,
```

async-openai/src/error.rs
```diff
@@ -109,6 +110,26 @@ impl std::fmt::Display for ApiError {
...
 
+/// `ApiError` paired with the HTTP status code from the response.
+#[cfg(feature = "_api")]
+#[derive(Debug, Clone)]
+pub struct ApiErrorResponse {
+    /// HTTP status code
+    pub status_code: reqwest::StatusCode,
+    /// Parsed error from response 
+    pub api_error: ApiError,
+}
```
64bit, line 121:
```suggestion
pub struct ApiErrorResponse {
    /// HTTP status code
    pub status_code: reqwest::StatusCode,
    /// Parsed error from response
    pub api_error: ApiError,
}
```

64bit (approved):
Thank you for the PR!

PR #554, opened 2026-05-20
fix(responses): default ResponseTextParam.format during deserialization

### LINKED ISSUE #551: Some fields are stricter then the openapi spec
The spec defines `ResponseTextParam` and `TextResponseFormatConfiguration` as

```yaml
ResponseTextParam:
  type: object
  description: |
    Configuration options for a text response from the model. Can be plain                                                                                                                                                                                                                                                                                                                                         
    text or structured JSON data. Learn more:                                                                                                                                                                                                                                                                                                                                                                      
    - [Text inputs and outputs](/docs/guides/text)                                                                                                                                                                                                                                                                                                                                                                 
    - [Structured Outputs](/docs/guides/structured-outputs)
  properties:
    format:
      $ref: '#/components/schemas/TextResponseFormatConfiguration'
    verbosity:
      $ref: '#/components/schemas/Verbosity'

TextResponseFormatConfiguration:
  description: >
    An object specifying the format that the model must output.
    
    
    Configuring `{ "type": "json_schema" }` enables Structured Outputs, 
    
    which ensures the model will match your supplied JSON schema. Learn more
    in the 
    
    [Structured Outputs guide](/docs/guides/structured-outputs).
    
    
    The default format is `{ "type": "text" }` with no additional options.
    
    
    **Not recommended for gpt-4o and newer models:**
    
    
    Setting to `{ "type": "json_object" }` enables the older JSON mode,
    which
    
    ensures the message the model generates is valid JSON. Using
    `json_schema`
    
    is preferred for models that support it.
  oneOf:
    - $ref: '#/components/schemas/ResponseFormatText'
    - $ref: '#/components/schemas/TextResponseFormatJsonSchema'
    - $ref: '#/components/schemas/ResponseFormatJsonObject'
```

The spec does not require `format`, but the definition of [`ResponseTextParam`](https://github.com/64bit/async-openai/blob/main/async-openai/src/types/responses/response.rs#L1128), requires it and `TextResponseFormatConfiguration` has no default. A more spec compliant definition would define `TextResponseFormatConfiguration::Text` as default, and `ResponseTextParam.format` would use the `#[serde(default)]` annotion.

Paul-E:
Closes #551

### DIFF
--- a/async-openai/src/types/responses/response.rs
+++ b/async-openai/src/types/responses/response.rs
@@ -1139,6 +1139,7 @@ pub struct ResponseTextParam {
     /// Setting to `{ "type": "json_object" }` enables the older JSON mode, which
     /// ensures the message the model generates is valid JSON. Using `json_schema`
     /// is preferred for models that support it.
+    #[serde(default)]
     pub format: TextResponseFormatConfiguration,
 
     /// Constrains the verbosity of the model's response. Lower values will result in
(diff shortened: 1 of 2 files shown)


64bit (approved):
Thank you!

PR #556, opened 2026-05-21
feat: forward reqwest's rustls-no-provider feature

### LINKED ISSUE #547: Allow reqwest 0.13's `rustls-no-provider` feature
The update to `reqwest` 0.13 causes some trouble if you want to rely on `ring` instead of `aws-ls-rs`.

See https://github.com/seanmonstar/reqwest/issues/2924. You might also want to reach out here: https://github.com/seanmonstar/reqwest/issues/2871 

I am still not sure what the best way for libraries is with `reqwest` 0.13. Nevertheless, unless there is a better and documented recommendation, async-openai should offer a way to enable `reqwest/rustls-no-provider` instead of `reqwest/rustls`

valkum:
This closes https://github.com/64bit/async-openai/issues/547

This adds a new feature `rustls-no-provider` which will forward to reqwest. This allows users to provide their own rustls crypto provider.

This contains additions to the CI jobs to prevent regressions in the future.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -118,3 +118,55 @@ jobs:
         env:
           RUSTFLAGS: "-D warnings"
         run: cargo build --no-default-features --features ${{ matrix.feature }},middleware --verbose
+
+  tls-features:
+    name: TLS Feature - ${{ matrix.feature }}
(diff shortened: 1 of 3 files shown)


64bit (approved):
Thank you for this, love it! 🎉

PR #558, opened 2026-05-21
Surface streaming HTTP errors at the function call instead of as first stream item

### LINKED ISSUE #553: Title: create_stream* methods return Ok(stream) on HTTP error; error surfaces as first stream item instead of as Err from the call
## Summary

  Streaming methods (`post_stream`, `post_stream_mapped_raw_events`, `get_stream`) return `Ok(stream)` even when the HTTP response status is non-2xx. The `OpenAIError::ApiError` is then delivered as the **first item of the returned stream**, not as the `Err` of the calling function. Callers that want HTTP failures surfaced as function-level errors (the natural Rust shape: `Result<Stream, Error>`) have to peek the first item themselves and prepend it back if it's a valid chunk.

  ## Where it happens

  `execute_stream` ([`src/client.rs#L638-L647`](https://github.com/64bit/async-openai/blob/main/async-openai/src/client.rs)) calls `execute_response().await?` and then unconditionally constructs the stream, regardless of status:

```rust
  async fn execute_stream<O>(...) -> Result<StreamResponse<O>, OpenAIError> {
      let response = self.execute_response(request_factory).await?;
      Ok(stream(response).await)   // <-- always Ok, even for 4xx/5xx
  }
```

  The status check is deferred into stream_mapped_raw_events, which on a bad status emits the error as the first stream item:

  - wasm path (src/client.rs#L775-L784):
```rust
  if !response.status().is_success() {
      return Box::pin(futures::stream::once(async move {
          match read_response(response).await {
              Ok(_) => Err(OpenAIError::InvalidArgument(...)),
              Err(error) => Err(error),
          }
      }));
  }
```
  - non-wasm path (src/client.rs#L831-L839):
```rust
  tokio::spawn(async move {
      if !response.status().is_success() {
          if let Err(e) = read_response(response).await {
              let _ = tx.send(Err(e));
          }
          return;
      }
      ...
  });
```

  ## Reproducer (conceptual)

  Any streaming call against a server that returns 400/401/etc.:

```rust
  let stream_result = client.chat().create_stream(bad_request).await;
  // stream_result is Ok(_), not Err(OpenAIError::ApiError(...))
  let mut stream = stream_result.unwrap();
  let first = stream.next().await;
  // first is Some(Err(OpenAIError::ApiError(...)))
```

  ## Why this matters

  For gateway / proxy callers (and any code that wants to short-circuit on HTTP failure before consuming a stream), the current shape forces a peek-and-prepend workaround at every call site:

```rust
  match stream.next().await {
      Some(Err(e)) => Err(e),
      Some(Ok(first)) => Ok(Box::pin(stream::once(async move { Ok(first) }).chain(stream))),
      None => Err(empty_stream_error()),
  }
```

  This is also surprising for new users — the natural Rust expectation is that Result<Stream, Error> distinguishes "couldn't open the stream" from "stream item failed." Today both collapse to Ok(stream_that_yields_err).

  ## Proposed change

  Check the response status inside execute_stream / execute_stream_mapped_raw_events and return Err(...) directly when non-2xx, before constructing the stream. The body parsing logic already exists in read_response (src/client.rs#L726-L752) — it just needs to be invoked from the streaming path on the non-2xx branch.

  ## Rough sketch:

```rust
  async fn execute_stream<O>(...) -> Result<StreamResponse<O>, OpenAIError> {
      let response = self.execute_response(request_factory).await?;
      if !response.status().is_success() {
          // Reuse read_response — already handles both JSON-bodied and raw 5xx error paths.
          let _ = read_response(response).await?;
          return Err(OpenAIError::InvalidArgument(
              "stream request failed without an error body".into(),
          ));
      }
      Ok(stream(response).await)
  }
```

  This is a breaking change for any caller that today relies on the error arriving as a stream item, but the new shape is strictly more informative (you can still get the item-style error by collecting from the stream if you want; the function-level error is the new option).

  Happy to send a PR if there's agreement on the approach.

max-zheng:
Surfaces non-2xx HTTP responses as `Err(OpenAIError::ApiError(_))` from the streaming methods (`post_stream`, `post_stream_mapped_raw_events`, `get_stream`) instead of delivering them as the first item of the returned stream. Centralizes the status check in `execute_response` so all callers (unary and streaming) share one code path for "non-2xx → error."

  Closes #553.

  ## Changes

  - `execute_response` now bails with `read_error_response(response).await` on non-success status, so every downstream caller sees only 2xx responses.
  - New `read_error_response` helper consumes a non-2xx response and produces the matching `OpenAIError::ApiError(ApiErrorResponse { .. })`. It's the single source of truth for HTTP error construction — previously this logic was duplicated inline in `read_response` and (incorrectly, via an `InvalidArgument` fallback) in the wasm `stream_mapped_raw_events`.
  - `read_response`, `execute_stream`, `execute_stream_mapped_raw_events`, and both `stream_mapped_raw_events` impls drop their now-redundant local status checks.
  - Removes the `OpenAIError::InvalidArgument(...)` fallback from the wasm streaming path (per maintainer feedback on #553 — `InvalidArgument` is reserved for client-side validation).

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -632,7 +632,11 @@ impl<C: Config> Client<C> {
         &self,
         request_factory: HttpRequestFactory,
     ) -> Result<Response, OpenAIError> {
-        self.executor.execute(request_factory).await
+        let response = self.executor.execute(request_factory).await?;
+        if !response.status().is_success() {
+            return Err(read_error_response(response).await);
+        }
+        Ok(response)
     }
 
     async fn execute_stream<O>(
@@ -724,15 +728,23 @@ impl<C: Config> Client<C> {
 }
 
 async fn read_response(response: Response) -> Result<(Bytes, HeaderMap), OpenAIError> {
-    let status = response.status();
     let headers = response.headers().clone();
     let bytes = response.bytes().await.map_err(OpenAIError::Reqwest)?;
+    Ok((bytes, headers))
+}
+
+async fn read_error_response(response: Response) -> OpenAIError {
+    let status = response.status();
+    let bytes = match response.bytes().await {
+        Ok(b) => b,
+        Err(e) => return OpenAIError::Reqwest(e),
+    };
 
     if status.is_server_error() {
         // OpenAI does not guarantee server errors are returned as JSON so we cannot deserialize them.
         let message: String = String::from_utf8_lossy(&bytes).into_owned();
         tracing::warn!("Server error: {status} - {message}");
-        return Err(OpenAIError::ApiError(ApiErrorResponse {
+        return OpenAIError::ApiError(ApiErrorResponse {
             status_code: status,
             api_error: ApiError {
                 message,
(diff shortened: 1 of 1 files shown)


64bit (approved):
This is so nice, thank you!

PR #561, opened 2026-06-01
fix(chat): skip_serializing_if on FunctionCallStream optional fields

### LINKED ISSUE #560: FunctionCallStream serializes None fields as JSON null instead of omitting them
[`FunctionCallStream`](https://github.com/64bit/async-openai/blob/main/async-openai/src/types/chat/chat_.rs#L1108-L1117) does not apply `#[serde(skip_serializing_if = "Option::is_none")]` to its two `Option<String>` fields (`name` and `arguments`), so a value constructed with `None` serializes as `"name": null` / `"arguments": null` rather than the field being absent.

This is inconsistent with the rest of `chat_.rs`, where neighbouring `Option<…>` fields use the attribute — e.g. [`service_tier`](https://github.com/64bit/async-openai/blob/main/async-openai/src/types/chat/chat_.rs#L1088) and [`system_fingerprint`](https://github.com/64bit/async-openai/blob/main/async-openai/src/types/chat/chat_.rs#L1092) a few lines earlier. 64 of the ~90 `Option<…>` fields in the file already use the attribute.

### Impact

Downstream OpenAI-shape consumers that read `function.arguments` with a string default (e.g. `dict.get('arguments', '')` in Python) get `None`/`null` back instead of `''`, and follow-up operations like `len(value)` or string concatenation then raise `TypeError`. The OpenAI streaming reference emits `""` on the first tool_call chunk and the [streaming-events schema](https://developers.openai.com/api/reference/resources/chat/subresources/completions/streaming-events) presents `arguments` as `optional string` — either an empty string or an absent field is wire-conformant; explicit `null` is not.

### Proposed fix

Add `#[serde(skip_serializing_if = "Option::is_none")]` to both fields. Two-line patch + a parallel test in `tests/ser_de.rs` mirroring the existing [`stream_options_none_fields_not_serialized`](https://github.com/64bit/async-openai/blob/main/async-openai/tests/ser_de.rs#L30).

### Related (downstream)

This is the upstream root cause for agentgateway/agentgateway#1988 (originally proposed locally as agentgateway/agentgateway#1989, closed in favour of fixing here).

pennycoders:
Closes #560.

The streaming `FunctionCallStream`'s `name` and `arguments` fields lacked the `#[serde(skip_serializing_if = "Option::is_none")]` attribute that the rest of `chat_.rs` uses on its `Option<…>` fields (64 of ~90 use it, including neighbours `service_tier` and `system_fingerprint`). When constructed with `None`, those fields serialized as JSON `null` rather than being absent, which breaks strict-string downstream accumulators that read the field with a string default.

## Changes
- `async-openai/src/types/chat/chat_.rs`: add `#[serde(skip_serializing_if = "Option::is_none")]` to `FunctionCallStream::name` and `FunctionCallStream::arguments`.
- `async-openai/tests/ser_de.rs`: new test `function_call_stream_none_fields_not_serialized` mirroring the existing `stream_options_none_fields_not_serialized` — covers field-omission, the empty-`{}` case, and round-trip deserialization.

## Verification
- `cargo fmt --all -- --check`: clean.
- `cargo check --all-features`: clean.
- `cargo test --test ser_de --all-features`: **3 / 3 passed** (including the new test).
- `cargo test --test responses_input_item_serde --all-features`: **6 / 6 passed**.
- `cargo test --test responses_output_to_input --all-features`: **6 / 6 passed**.
- The remaining test failures observed locally (`embedding::tests::*`, `file::tests::*`, `vectorstores::vector_store_files::tests::*`, `boxed_future_test`) are pre-existing on `main` — they instantiate `Client::new()` and exercise the live OpenAI API, so they fail without `OPENAI_API_KEY` set. Verified identical failures on `main` without this patch. Not introduced by this change.

## Related (downstream)

This is the upstream root cause for agentgateway/agentgateway#1988 (originally proposed locally as agentgateway/agentgateway#1989, closed in favour of fixing here).


64bit (approved):
Thanks for context and PR, LGTM!

PR #535, opened 2026-04-07
Allow missing id in ReasoningItem

howardjohn:
While this is a required field in the OpenAPI spec, the spec is wrong; OpenAI itself accepts a missing `id` and popular tools like OpenCode and Codex send requests without an `id`. This aligns the behavior with the real world usage to allow deserializing OpenCode/Codex (and other) requests

PR #570, opened 2026-07-02
fix(responses): make OutputMessage.id optional (fixes #565)

### LINKED ISSUE #565: `CreateResponses` requires `id` field when OpenAI itself does not require it
With a request:

```json
{
    "model": "gpt-5.5",
    "input": [
        {
            "content": [
                {
                    "text": "hello",
                    "type": "output_text"
                }
            ],
            "role": "assistant",
            "type": "message"
        },
        {
            "content": [
                {
                    "text": "what did you say to me last?",
                    "type": "input_text"
                }
            ],
            "role": "user",
            "type": "message"
        }
    ]
}
```

 we cannot deserialize. This same content is accept by openai.com and sent via codex. The issue is a missing `id` on the `output_text`; the SDK currently treats it as required.

While the JSON schema does say ID is required it appears to not be in practice with the official API and official client.

fuleinist:
The OpenAI Responses API sometimes omits the `id` field on `output_text` items when they appear in the input array — for example, multi-turn conversations passing assistant messages back to the model. Previously `OutputMessage.id` was `String` (required), causing deserialization failures like:

```
CreateResponses requires id field when OpenAI itself does not require it
```

Changed to `Option<String>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — the same pattern already used by `ReasoningItem.id` (PR #535).

Closes #565


commit_msg.txt
```diff
@@ -0,0 +1,12 @@
+fix(responses): make OutputMessage.id optional
```
howardjohn: should prob delete this?

howardjohn (approved):
LGTM personally

commit_msg.txt
fuleinist: Good catch — deleted the stray commit_msg.txt in 3ba253b5a00eda9c1fe2f8c97fe624f73c8aa1b4: https://github.com/64bit/async-openai/commit/3ba253b5a00eda9c1fe2f8c97fe624f73c8aa1b4. Thanks for the review!

64bit (changes requested):
Thanks @fuleinist for the PR - to provide you context - when spec is mismatched from reality it requires to have a test or an example to validate that deviation still works in practice. 

> please include a test to validate with OpenAI API.

While this is simple change, how do I verify that the claim is correct and maintain it? Hence I require you to add test or an example which exercises the change with real API.

fuleinist:
@64bit addressed. Added two real API integration tests in \
esponses_output_message_id.rs\:

1. **\output_message_id_round_trip\** — multi-turn pattern: creates a response, echoes the full \output\ back as \input\ (the canonical fix for #565), then makes a second request. Verifies the round-trip deserializes without error.

2. **\output_message_without_id_in_input\** — constructs an \InputItem\ from an \OutputMessage\ that deliberately omits \id\ (simulating what the API sometimes returns), sends it to the API, and confirms no deserialization panic.

Both tests skip gracefully when \OPENAI_API_KEY\ is unset. Also registered the existing \
esponses_output_to_input\ test in \Cargo.toml\ (was missing from \[[test]]\ sections).

Unit tests pass (14/14):
\\\
test result: ok. 8 passed (responses_input_item_serde)
test result: ok. 6 passed (responses_output_to_input)
\\\

Integration test compiles cleanly with zero warnings.

commit_msg.txt
fuleinist: Done — commit_msg.txt was removed in 3ba253b (chore: remove stray commit_msg.txt from PR diff). New commit 24a91851d07351fa203b1616084daa5c4885b3f8: https://github.com/64bit/async-openai/commit/24a91851d07351fa203b1616084daa5c4885b3f8

fuleinist:
Thanks for the review context — addressed.

This branch now ships two layers of API-exercising coverage for the OutputMessage.id optional change:

1. **Integration tests** (commit 28177a0): output_message_id_round_trip and output_message_without_id_in_input hit the real OpenAI Responses API in async-openai/tests/responses_output_message_id.rs. They skip automatically when OPENAI_API_KEY is unset.

2. **Runnable example** (commit 24a91851d07351fa203b1616084daa5c4885b3f8: https://github.com/64bit/async-openai/commit/24a91851d07351fa203b1616084daa5c4885b3f8): examples/responses-multi-turn-output-message demonstrates the multi-turn round-trip — turn 1 captures the assistant OutputMessage, the example echoes it back as input via MessageItem::Output, turn 2 asks the model to follow up. This is exactly the path the fix protects.

`OPENAI_API_KEY=... cargo run -p responses-multi-turn-output-message` to try it.

Would appreciate a re-review when you have a moment.

fuleinist:
@64bit Thanks for the review — addressed in commits `28177a0e` (integration test against the real OpenAI API) and `24a91851` (multi-turn round-trip example). 

Per your request "a test or example which exercises the change with real API":

1. **Integration test** `async-openai/tests/responses_output_message_id.rs` — calls the real `POST /v1/responses` endpoint with `OPENAI_API_KEY`, then echoes the resulting `OutputMessage` back as input for a second turn to verify it serializes/deserializes without error (skipped when the key is absent so CI doesn't break).

2. **Example** `examples/responses-multi-turn-output-message/` — runnable with `OPENAI_API_KEY=… cargo run -p responses-multi-turn-output-message`; does a full multi-turn round-trip using `Item::MessageItem(MessageItem::Output(...))`.

3. **Unit tests** in `async-openai/tests/responses_input_item_serde.rs` — `output_message_without_id_deserializes` and `output_message_with_id_still_works` cover both the deserialization and `skip_serializing_if` behaviour without needing network access.

Could you take another look when you have a moment?

fuleinist:
Thanks for the review @64bit — I've added both an integration test and a runnable example that exercise the change against the real OpenAI API:

1. **Integration test** — `async-openai/tests/responses_output_message_id.rs` (commit 28177a0):
   - `output_message_id_round_trip`: turn-1 → echo `response.output` back as input → turn-2 succeeds (validates deserialization of any real-world OutputMessage)
   - `output_message_without_id_in_input`: builds an input with an OutputMessage that has no `id` field and confirms the API accepts it
   - Both gate on `OPENAI_API_KEY` and self-skip when absent

2. **Runnable example** — `examples/responses-multi-turn-output-message/` (commit 24a91851):
   - `cargo run -p responses-multi-turn-output-message` does a real multi-turn conversation and prints which OutputMessage had `id` set vs `None`
   - Self-documenting Cargo.toml doc-comment explains the multi-turn pattern

HEAD: `24a91851d07351fa203b1616084daa5c4885b3f8` → https://github.com/64bit/async-openai/commit/24a91851d07351fa203b1616084daa5c4885b3f8

Re-requesting your review — happy to iterate if either the test surface or the example needs adjustments.

howardjohn:
@64bit I am interested in this, wasn't sure what the close was for and if I could open a new PR resolving those issues or if this won't be accepted. Thanks!

PR #571, opened 2026-07-07
Add Atlas Cloud chat example

binyangzhu000-sudo:
## Summary
- add a minimal Atlas Cloud chat example using async-openai
- configure the OpenAI-compatible Atlas Cloud base URL with `ATLASCLOUD_API_KEY`
- allow overriding the default live Atlas text model through `ATLASCLOUD_MODEL`

## Validation
- `rustfmt --edition 2021 examples/atlascloud-chat/src/main.rs`
- `cargo check -p atlascloud-chat`

### DIFF
--- a/examples/atlascloud-chat/Cargo.toml
+++ b/examples/atlascloud-chat/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "atlascloud-chat"
+version = "0.1.0"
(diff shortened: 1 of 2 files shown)


64bit:
Thanks! but its not doing anything new that the existing examples don't already cover.

PR #572, opened 2026-07-10
feat: upstream spec sync

howardjohn:
Following the standard flow for the new sync. They seemed to change the openapi spec substantially as part of some structure change, hence the massive yaml diff. I spot checked all of the chat and responses changes and they look correct based on recent additions (explicit cache points, inline moderation, and `Max` reasoning)


64bit:
Please see https://github.com/openai/openai-openapi/issues/557

howardjohn:
@64bit thanks! I updated back to the github repo. Still failing a few test but working on it.

async-openai/src/types/shared/reasoning_effort.rs
```diff
@@ -10,4 +10,5 @@ pub enum ReasoningEffort {
     Medium,
     High,
     Xhigh,
+    Max,
```
michaelfeil, line 13: this would be useful for me.

PR #577, opened 2026-07-30
Exit stream reader task on consumer drop

### LINKED ISSUE #576: Streaming reader task leaks when the consumer is dropped mid-stream (stream_mapped_raw_events never observes receiver drop)
### Streaming reader task leaks when the consumer is dropped mid-stream

  In `stream_mapped_raw_events` (`src/client.rs`), the spawned task that drains the response only notices a dropped consumer via `tx.send(...).is_err()` — i.e. *after* the next event arrives:

  ```rust
  tokio::spawn(async move {
      while let Some(ev) = event_stream.next().await {   // parks here
          ...
          if tx.send(response).is_err() { break; }        // only disconnect check
      }
  });
  ```

  If the returned stream is dropped (client disconnect / cancellation) while the task is parked at `event_stream.next().await` and the upstream is open-but-idle (no more chunks, no `[DONE]`, no error), the task never reaches `tx.send`, never
  observes the dropped receiver, and **parks indefinitely** — holding the `reqwest::Response`, connection, and buffers. In a long-lived server proxying streams, these orphaned tasks accumulate: RSS grows monotonically with abandoned streams,
  decoupled from CPU. The detached `JoinHandle` and lack of an intrinsic timeout mean nothing else can cancel it.

  (The wasm path uses pull-based `stream::unfold` and doesn't have this issue.)

  **Suggested fix:** also observe the receiver dropping, so the task exits immediately on consumer drop regardless of upstream state:

  ```rust
  loop {
      tokio::select! {
          biased;
          _ = tx.closed() => break,
          maybe_ev = event_stream.next() => {
              let Some(ev) = maybe_ev else { break };
              // ... existing handling ...
              if tx.send(event_mapper(/* event */)).is_err() { break; }
          }
      }
  }
  ```

  `UnboundedSender::closed()` resolves as soon as the receiver drops, so the task tears down the upstream connection promptly without relying on a caller-configured timeout. Happy to open a PR.

max-zheng:
**Fix stream reader task leaking when the consumer is dropped**

  Closes #576.

  In `stream_mapped_raw_events` (non-wasm), the spawned reader task only noticed a dropped consumer via `tx.send(..).is_err()` — i.e. after the *next* event arrived. If the returned stream was dropped while the task was parked on `event_stream.next()` and upstream was open but idle (no more chunks, no `[DONE]`, no error), the task parked indefinitely, holding the `reqwest::Response`, its connection, and buffers. In a long-lived server proxying streams, these orphaned tasks accumulate and RSS grows with abandoned streams.

  The reader now also selects on `tx.closed()`, which resolves as soon as the receiver is dropped, so the task exits and releases the upstream response immediately regardless of upstream state. The loop body is unchanged; a closed consumer simply ends the existing `while let`.

  The wasm path is pull-based (`stream::unfold`) and is unaffected.

### DIFF
--- a/async-openai/src/client.rs
+++ b/async-openai/src/client.rs
@@ -839,7 +839,15 @@ where
             .map(|r| r.map_err(std::io::Error::other));
         let mut event_stream = std::pin::pin!(eventsource_stream::EventStream::new(byte_stream));
 
-        while let Some(ev) = event_stream.next().await {
+        // Also observe the consumer dropping the stream: relying on
+        // `tx.send(..).is_err()` alone would keep this task - and the upstream
+        // response it holds - alive until the next event arrives, which never
+        // happens when upstream is open but idle.
+        while let Some(ev) = tokio::select! {
+            biased;
+            _ = tx.closed() => None,
+            ev = event_stream.next() => ev,
+        } {
             let event = match ev {
                 Ok(e) => e,
                 Err(e) => {
@@ -871,7 +879,7 @@ where
 #[cfg(all(test, feature = "middleware", not(target_family = "wasm")))]
 mod tests {
     use std::sync::{
-        atomic::{AtomicUsize, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc,
     };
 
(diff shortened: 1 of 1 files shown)


64bit (approved):
Thank you, LGTM


Looks like CI failures are unrelated and looks new I'll take care of that.
https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#useless_borrows_in_formatting

PR #584, opened 2026-08-13
fix(responses): preserve Codex client metadata

Jeadie:
OpenAI Codex includes request metadata in `v1/responses` under `client_metadata`.

For example:

```json
{
  "client_metadata": {
    "x-codex-installation-id": "...",
    "thread_id": "...",
    "session_id": "...",
    "x-codex-window-id": "...",
    "turn_id": "...",
    "x-codex-turn-metadata": "..."
  }
}
```

Preserve this metadata in the typed Responses request so gateways can forward it. Add a serde roundtrip test covering all Codex metadata fields.

This is backwards compatible: the field is optional, and requests that do not include Codex metadata are unchanged.


Jeadie:
Superseded by #585, which combines this with the other Codex Responses PR and gates both behind an off-by-default `codex` feature.

PR #583, opened 2026-08-13
fix(responses): support Codex additional tools

Jeadie:
OpenAI Codex has a new input item for `v1/responses` (type `additional_tools`). 
```json
  {
    "type": "additional_tools",
    "role": "developer",
    "tools": [ ... ]
  }
```
Update `v1/responses` types accordingly, add roundtrip test. Backwards compatible, net new variant.

PR #585, opened 2026-08-15
feat(responses): add Codex request variants behind a `codex` feature

Jeadie:
## Summary

OpenAI Codex sends two extras on `v1/responses` that the current Responses types drop. This adds both, behind a new off-by-default `codex` feature.

- `additional_tools` input item (type `additional_tools`) — carries client tool definitions.
- `client_metadata` on the request — carries Codex session identifiers.

A gateway that sits in front of Codex deserializes the request into `CreateResponse` and serializes it again before it forwards the request. Without these types, the two extras are lost, so the forwarded request differs from what Codex sent.

## Feature gate

Both additions are Codex-specific variants on the input and request types. They are gated behind a new `codex` feature (implies `response-types`) and are off by default. A build that does not target Codex keeps the smaller type surface. A build that serves or proxies Codex enables `codex`.

```toml
async-openai = { version = "*", features = ["responses", "codex"] }
```

## Supersedes

Combines and replaces #583 (`additional_tools`) and #584 (`client_metadata`) into one PR, and adds the feature gate the two did not have.

## Compatibility

Additive and backwards compatible. With `codex` off, the types are absent and the build is unchanged. With `codex` on, both fields are optional, so a request without them still round-trips.

## Validation

- `cargo build -p async-openai --no-default-features --features response-types` — builds with the variants compiled out.
- `cargo test -p async-openai --no-default-features --features response-types,codex --test responses_input_item_serde` — round-trip tests for both variants pass; they are skipped without `codex`.
- `cargo fmt --check` and `cargo clippy --features response-types,codex -- -D warnings` are clean.


64bit (changes requested):
Is this officially documented anywhere? if so please add doc links.

If its not documented API, please add a test which make real API call to verify what is claimed in the PR, PR should not be testing serde in tests.

